PR #24, opened 2025-01-27
[TSK-797] Implement course metadata model for the course card component

### LINKED ISSUE #10: Implement the CourseMetadata core model
In particular, to be used in the course card UI kit component.

alebg:
Fix #10

### DIFF
--- a/packages/models/src/course.ts
+++ b/packages/models/src/course.ts
@@ -1,9 +1,34 @@
 import { z } from 'zod'
+import { LanguageSchema } from './language'
 
-export const CourseMetedataSchema = z.object({
+
+export const CourseDurationSchema = z.object({
+    video: z.number(),  // in minutes, duration of video content
+    coaching: z.number(),  // in minutes, duration of coaching sessions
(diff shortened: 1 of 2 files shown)


packages/models/src/course.ts
```diff
@@ -1,9 +1,32 @@
...
+
+export const CourseMetadataSchema = z.object({
     title: z.string(),
     description: z.string(),
-    duration: z.number(),
+    duration: CourseDurationSchema,
+    pricing: CoursePricingSchema,
+    imageUrl: z.string(),
+    rating: z.number(),
+    author: z.string(),
```
alebg, line 27: @maany remark here: so, this more or less implies that we'll send from the backend these two profile-related fields (author and authorImage). Is that OK? As opposed to putting here a whole Profile object

maany: Sure, just a minor comment, you can have 
```
author: {
  name: z.string(),
  image: z.string()
}
```

maany (changes requested):
author profile refactoring

Issue #2, opened 2025-01-02
Configure tailwind `@source` directive for production environments in nextjs applications 


maany:
The `global.css` file in NextJS 15 apps configures the `@source` directive for tailwindcss 4.0 to apply tailwind styles to ui-kit. This is currently being done just for development mode.

In production, the NextJS app will require setting the `source` directive to wherever the node_modules are present and must point to the dist/css/tailwind.css file.

alebg:
@maany what's the status of this issue?

maany:
We will need to fix this very soon, especially once we start building CMS pages

PR #25, opened 2025-01-28
Issue 17 configure tailwind css and i18n

### LINKED ISSUE #17: Configuration of tailwind css and internalization i18n for production environments in nextjs
The global.css file in NextJS 15 apps configures the @source directive for tailwindcss 4.0 to apply tailwind styles to ui-kit. This is currently being done just for development mode.

The middleware file have to be created for internalization i18n so that it can be adapted to various languages and regions without engineering changes. 

In production, the NextJS app will require setting the source directive to wherever the node_modules are present and must point to the dist/css/tailwind.css file.

Obitodevvv:
fixed #17


apps/platform/src/components/home.tsx
```diff
@@ -9,15 +9,14 @@ export interface HomeProps {
...
+    <div className="flex flex-col  text-base-neutral-50 gap-4 mt-3 items-center justify-center text-center">
       <p className="text-center">Cool things happen here</p>
       <Button
         variant="primary"
         size="medium"
         onClick={() => {
           console.log('Clicked: ' + theme);
-          setTheme(theme === 'orange' ? 'blue' : 'orange')
-
+          setTheme(theme === 'just-do-add' ? 'Job-rand-me' : 'Bewerbeagenture')
```
maany, line 19: name: Bewerbeagentur

apps/platform/src/middleware.ts
```diff
@@ -0,0 +1,9 @@
+import createMiddleware from 'next-intl/middleware';
+import { routing } from './i18n/routing';
+
+export default createMiddleware(routing);
+console.log("hello");
```
maany, line 5: I understand this is required for debugging, but please add a todo to remove this log and remove it once it is no longer needed. This could explode the prod logs

apps/platform/jest.config.ts
```diff
@@ -1,12 +0,0 @@
-const jestConfig = {
```
maany, line 1: why was this removed?

tsconfig.base.json
```diff
@@ -16,7 +16,7 @@
     "baseUrl": ".",
     "paths": {
       "@dad-e-class/dad-nx-vite": ["tools/dad-nx-vite/src/index.ts"],
-      "@dad-e-class/models": ["packages/models/src/index.ts"],
+      "@maany_shr/e-class-models": ["packages/models/src/index.ts"],
       "@dad-e-class/nx-dad": ["tools/nx-dad/src/index.ts"],
       "@maany_shr/e-class-ui-kit": ["packages/ui-kit/lib/index.ts"],
```
maany, line 21: See this typescript path mapping. It allows us to not be dependent on published ui-kit releases when developing the nextjs apps.

package.json
```diff
@@ -8,8 +8,11 @@
   },
   "private": true,
   "dependencies": {
+    "@maany_shr/e-class-models": "^1.4.2",
```
maany, line 11: adding this here would make the entire project dependent on the published npm package and not the packages/ui-kit directory. We have a typescript path config that circumvents checking for @maany_shr/e-class-models in the node_modules folder and instead maps all paths to the packages/ui-kit folder.

Is this required?

tsconfig.base.json
```diff
@@ -16,7 +16,7 @@
     "baseUrl": ".",
     "paths": {
       "@dad-e-class/dad-nx-vite": ["tools/dad-nx-vite/src/index.ts"],
-      "@dad-e-class/models": ["packages/models/src/index.ts"],
+      "@maany_shr/e-class-models": ["packages/models/src/index.ts"],
```
maany, line 19: Aha, good catch!

maany (changes requested):
Need to address some queries about the project setup

maany:
@alebg can you add this PR to a milestone on the right column :)

apps/platform/jest.config.ts
Obitodevvv: Because jest testing we have to do for components so we are doing in the ui-kit. And playwright testing we are doing in platform-e2e . Similarly for storybook we are done in ui-kit. From my opinion this jestConfig is not required here in this.

tsconfig.base.json
Obitodevvv: So, here we need to add * like this ? - "@maany_shr/e-class-models":"*"

maany: No, this looks good. I was just making a reference to line 21 to indicate that the root package.json does not need to install @maany_shr/e-class-ui-kit since we have a typescript path mapping here :)

apps/platform/jest.config.ts
maany: If you have this file, does the `paltform:test` target fail as there are no tests?

package.json
Obitodevvv: This is our main doubt that - is it needed to add all packages in the root packages.json file ?

apps/platform/jest.config.ts
maany: playwright tests are quite expensive to run as they spin up real browsers (multiple instances). To test external API connections and not the entire e2e workflow for each api call, it might be useful to use jest in the node runtime and make sure we can handle the responses correctly. So, even though we are not using jest in the platform at the moment, it may come in handy later., :)

package.json
maany: The packages which are in the monorepo need a typescript path alias. The root package.json should contain only the externally required packages ( available on npm).

apps/platform/jest.config.ts
Obitodevvv: Yes, we got issues while targeting tests in the platform:test . The issue we are facing in  themeprovider -"Cannot read properties of null (reading 'useState')".

package.json
maany: # Managing Dependencies in a Monorepo

## Packages in a Monorepo as External Dependencies  
When publishing a package in your monorepo to a public registry (like npm), it needs to properly declare its dependencies. For example:  

- **ui-kit** is a package in the monorepo that we publish to npm.  
- **ui-kit** depends on the library **lucide-react**.  

To ensure this dependency works as expected:  
1. Add **lucide-react** to the root `package.json` of the monorepo. This is required because the monorepo centralizes dependency management.  
2. Also, add **lucide-react** to the `package.json` of **ui-kit**. This ensures that when **ui-kit** is installed as a dependency in another project, its required dependencies (like **lucide-react**) are correctly downloaded from npm.

**Key point**: Packages published to npm do not retain any special knowledge of being part of a monorepo. Each package must explicitly declare its own dependencies for them to work correctly outside the monorepo.

---

## Packages in a Monorepo as Internal Dependencies  
When one monorepo package depends on another (but is not being published externally):  

- **platform** is a project in the monorepo that uses **ui-kit** as an internal dependency.  

In this case, you don’t need to publish **ui-kit** or worry about external dependency declarations. Instead:  
1. Define a TypeScript path alias in the root `tsconfig.json` file to link **platform** to **ui-kit**.  
   - Example: `"@ui-kit/*": ["packages/ui-kit/src/*"]`  
2. The alias ensures that **platform** can directly reference and use **ui-kit** during development without needing to install it separately.

**Key point**: Internal dependencies leverage TypeScript path aliases for easy access and don’t require npm publishing or dependency declarations.

apps/platform/jest.config.ts
Obitodevvv: Sir, if we are using themeprovider in the testing. Then we facing issue. I guess some issues are there with config.

apps/platform/src/i18n/routing.ts
```diff
@@ -0,0 +1,15 @@
+import {defineRouting} from 'next-intl/routing';
+import {createNavigation} from 'next-intl/navigation';
+ 
+export const routing = defineRouting({
+  // A list of all locales that are supported
+  locales: ['en', 'es','fr'],
```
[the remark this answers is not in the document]
alebg: @Obitodevvv this is still pending

apps/platform/jest.config.ts
maany: Roger that! Since the current main branch has the platform:test target passing properly, I believe the issue is coming from the fact that everything in platform now lives under the [locale] directory and jest is not able to resolve imports due to special characters. I can take a look at this in a separate issue

maany: This would need further investigation in #27

maany:
@alebg I have reviewed the changes and things form the repo setup look okay. Please review the rest.

maany:
p.s this does not resolve #2

alebg:
> p.s this does not resolve #2. That will be handled later when we build the prod containers for platform and cms

@Obitodevvv please remove "and #2" from your PR message, otherwise that issue will be closed when we merge this

alebg (changes requested):
@Obitodevvv looks good also to me. Please update the locales as requested and then we're ready to merge this

apps/platform/src/i18n/routing.ts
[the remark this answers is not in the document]
Obitodevvv: @alebg  Now its done sir. Added 'de' and remove - 'es' and 'fr'. Now we have two lang - 'en' and 'de'.

alebg (approved):

alebg (changes requested):
@Obitodevvv this branch was not up to date with the main branch and has conflicts. Please rebase and fix the conflicts, otherwise we cannot merge

alebg:
> > @Obitodevvv this branch was not up to date with the main branch and has conflicts. Please rebase and fix the conflicts, otherwise we cannot merge
> 
> Sir, locally this branch is up-to-date with main branch already. Can you please provide me one screenshot of conflict ?

@Obitodevvv So, on approval, GitHub told me there were merge conflicts. Then, locally, if I run 

```
git checkout main
git pull
git checkout issue_17_configure_tailwindCSS_and_i18n
git pull
git rebase main
```

I see the conflicts. More concretely in the `packages/ui-kit/lib/components/professionalCard/FileDisplay.tsx` file for the first conflicting commit, but there could be other files.

alebg:
> > @Obitodevvv this branch was not up to date with the main branch and has conflicts. Please rebase and fix the conflicts, otherwise we cannot merge
> 
> Sir, locally this branch is up-to-date with main branch already. Can you please provide me one screenshot of conflict ?

@Obitodevvv It seems to me that the issue has to do with this branch's git history, which in previous commits included component files that then were deleted. For some reason these conflict in some past commit with the main branch's history. In particular, if I do, locally

```
git checkout main
git pull
git checkout issue_17_configure_tailwindCSS_and_i18n
git pull
git rebase main
```

At some point in the commit history I get the deleted files:
immagine
Plus a conflict on `FileDisplay.tsx`. Please run the rebase locally yourself as explained above and fix these conflicts so we can merge.

Obitodevvv:
> > > @Obitodevvv this branch was not up to date with the main branch and has conflicts. Please rebase and fix the conflicts, otherwise we cannot merge
> > 
> > 
> > Sir, locally this branch is up-to-date with main branch already. Can you please provide me one screenshot of conflict ?
> 
> @Obitodevvv So, on approval, GitHub told me there were merge conflicts. Then, locally, if I run
> 
> ```
> git checkout main
> git pull
> git checkout issue_17_configure_tailwindCSS_and_i18n
> git pull
> git rebase main
> ```
> 
> I see the conflicts. More concretely in the `packages/ui-kit/lib/components/professionalCard/FileDisplay.tsx` file for the first conflicting commit, but there could be other files.

Yes sir, I also uses this approach to solve this but here  `packages/ui-kit/lib/components/professionalCard/FileDisplay.tsx`  - these all files are in the different branch, not in the main branch or in this 'issue_17_configure_tailwindCSS_and_i18n' . These all files are the part of branch 'issue13_useprofile_compo' .  Only difference on main and 'issue_17_configure_tailwindCSS_and_i18n' branch is on jest.config.js file and spec/index.spec.tsx these two files are removed.

alebg:
> > > > @Obitodevvv this branch was not up to date with the main branch and has conflicts. Please rebase and fix the conflicts, otherwise we cannot merge
> > > 
> > > 
> > > Sir, locally this branch is up-to-date with main branch already. Can you please provide me one screenshot of conflict ?
> > 
> > 
> > @Obitodevvv So, on approval, GitHub told me there were merge conflicts. Then, locally, if I run
> > ```
> > git checkout main
> > git pull
> > git checkout issue_17_configure_tailwindCSS_and_i18n
> > git pull
> > git rebase main
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I see the conflicts. More concretely in the `packages/ui-kit/lib/components/professionalCard/FileDisplay.tsx` file for the first conflicting commit, but there could be other files.
> 
> Yes sir, I also uses this approach to solve this but here `packages/ui-kit/lib/components/professionalCard/FileDisplay.tsx` - these all files are in the different branch, not in the main branch or in this 'issue_17_configure_tailwindCSS_and_i18n' . These all files are the part of branch 'issue13_useprofile_compo' . Only difference on main and 'issue_17_configure_tailwindCSS_and_i18n' branch is on jest.config.js file and spec/index.spec.tsx these two files are removed.

@Obitodevvv The files were removed, yes, but they remain in the git history of this branch, 'issue_17_configure_tailwindCSS_and_i18n', as it is in this GitHub repo. See for example [this commit in the history of this branch](https://github.com/dream-aim-deliver/dad-e-class-monorepo/pull/25/commits/5505cb00158c190b6d557cea74dffa2d34e9cdb1) that explicitly introduced said files. If they were removed later in a more recent commit, it doesn't change the fact that they still live in a previous commit in the history. And if this history conflicts with the history of the main branch, then we cannot merge.

What you can do to replicate locally is clone this repo fresh elsewhere in your computer, and run 
```
git checkout issue_17_configure_tailwindCSS_and_i18n
git pull
git rebase main
```

Then you'll see the conflicts.

maany (dismissed):
Approving for the changes I had requested on the PR, they were addressed.

alebg:
> > > > @Obitodevvv this branch was not up to date with the main branch and has conflicts. Please rebase and fix the conflicts, otherwise we cannot merge
> > > 
> > > 
> > > Sir, locally this branch is up-to-date with main branch already. Can you please provide me one screenshot of conflict ?
> > 
> > 
> > @Obitodevvv So, on approval, GitHub told me there were merge conflicts. Then, locally, if I run
> > ```
> > git checkout main
> > git pull
> > git checkout issue_17_configure_tailwindCSS_and_i18n
> > git pull
> > git rebase main
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I see the conflicts. More concretely in the `packages/ui-kit/lib/components/professionalCard/FileDisplay.tsx` file for the first conflicting commit, but there could be other files.
> 
> Yes sir, I also uses this approach to solve this but here `packages/ui-kit/lib/components/professionalCard/FileDisplay.tsx` - these all files are in the different branch, not in the main branch or in this 'issue_17_configure_tailwindCSS_and_i18n' . These all files are the part of branch 'issue13_useprofile_compo' . Only difference on main and 'issue_17_configure_tailwindCSS_and_i18n' branch is on jest.config.js file and spec/index.spec.tsx these two files are removed.

@Obitodevvv on further thought, could you please squash all the commits of this PR into a single one? to prevent having an unruly git history. This may already solve the merge conflict, but if it does not, then please create a second merge commit

alebg (changes requested):
@Obitodevvv could you please squash all the commits of this PR into a single one? we want to prevent an unruly git history from the start. And please make the git commit message more descriptive of what you actually did

alebg (approved):
Great job @Obitodevvv , thanks a lot!

PR #33, opened 2025-02-02
Course-related Models Pt. 1

alebg:
Implemented models for package, category, and topic, with tests.

Named "EClassPackage" instead of just "Package" since the latter is a reserved word.


maany (approved):
LGTM!

PR #34, opened 2025-02-03
Remove rating from the course model

alebg:


maany:
Hey @alebg, let's quickly chat about this again. I know we discussed to calculate this value dynamically, however, after considering how frequently we might have to run these calculations and the fact they will only change during deterministic events ( like when a review is submitted), it might be easier on the db to just save it in the same course table.

PR #36, opened 2025-02-06
feat(i18n) : Implement centralized translation management.

### LINKED ISSUE #35: Implement centralized translation management for both platform app and ui-kit.
Need to create a shared translations package in the NX monorepo to centralize i18n logic for both the platform app and ui-kit library. 
Acceptance Criteria
- platform supports multiple languages via the shared package.
- ui-kit components render text based on the active locale & use translation keys.

ObitoDevv:
fix #35 ...


packages/translations/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "module": "commonjs",
```
maany, line 4: why commonjs?

packages/translations/src/lib/get-dictionary.ts
```diff
@@ -0,0 +1,44 @@
+import 'server-only';
```
maany, line 1: if the ui-kit is dependent on this, does it mean the components can be rendered only server side?

packages/ui-kit/lib/components/button.tsx
```diff
@@ -1,47 +1,143 @@
 "use client";
+import { FC, ReactNode, isValidElement } from 'react';
 import { cva, VariantProps } from 'class-variance-authority';
 import { cn } from '../utils/style-utils';
+import { useTranslations } from 'next-intl';
```
maany, line 5: This means the ui-kit is bound to NextJS. Also, the `package.json` of ui-kit would require a dependency to `next-intl` to be specified.

packages/translations/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "@dad-e-class/translations",
```
maany, line 2: if the ui-kit is published to npm, it will not bundle this project. So, if we use the ui-kit outside of this monorepo, translations will not work. ( for example, in the CMS)

maany (changes requested):
This needs discussion.

packages/ui-kit/lib/components/button.tsx
ObitoDevv: - Sir, are you suggesting that we should use 'react-i18next' instead of 'next-intl'?
- Or are you advising that we should import useTranslation from packages/translation rather than importing it from 'next-intl'?

packages/translations/package.json
ObitoDevv: So, @dad-e-class/translation should be installed in the ui-kit package.json ?

Like this - 
"dependencies": {
    "@dad-e-class/translations": "*",

alebg:
Closed in favor of #40

Issue #96, opened 2025-03-19
[TSK-785] models: implement all models related to HomePage


alebg:
Make a decision on how to capture all the data needed to feed the HomePage. E.g., implement the `Platform` model, and any other that may be needed.

alebg:
Fixed in #101

PR #101, opened 2025-03-22
[TSK-814] Add schemas and validation tests for home page structure

alebg:
Introduce schemas for the home page, including components like the banner, carousel, coaching on demand, and accordion list, along with validation tests to ensure proper structure and required fields.


alebg:
Fix #96

Issue #88, opened 2025-03-17
[TSK-816] models: implement CoachingSessionReview, instead of CoachReview; remove rating from CoachingSession


alebg:
- Rating will be in CoachingSessionReview
- There's no "CoachReview" in the abstract in figma, just coaching session reviews, and that's what gives a coach a rating

alebg:
Fixed in #102

PR #102, opened 2025-03-22
[TSK-816] Rename 'coach review' to 'coaching session review' and adjust schema

alebg:
Rename the coach review schema to coaching session review and move the rating field to the appropriate schema. Update tests to reflect these changes.


ydamit (approved):
Looks like exactly the right way to handle the review happening on the separate Coaching Review model. An individual Coaching Session should _not_ have an overall rating, since each of these will only get exactly one review anyway. That raises the question of whether the data for the review of a single Coaching Session is used in any way other than as part of calculating a Coach's (or Course's) overall rating. It's possible there would be some advantages to computing those roll-ups immediately each time a new review is submitted, and storing that on the Coach or Course model. But it sounded like there were other considerations to take into account before we can address that, so it should be in a separate task and PR, anyway.

alebg:
Fix #88

PR #111, opened 2025-03-24
[TSK-819] [TSK-820] ui-kit:Rich-text-element

### LINKED ISSUE #105: [TSK-819] [TSK-820] ui-kit: RichtextElement
We need a new subcomponent, `RichtextElement`, with two modes: **edit** and **view**. This will be used anywhere we need to edit or render richtext, using SlateJS.
  
- **Edit Mode:**  
  - Receives a Slate.js object created by the user.  
  - Converts it to a string before sending it out (e.g., via a callback).  

- **View Mode:**  
  - Accepts a string.  
  - Parses it with Slate.js and renders it.  

### Requirements 

- Implement this as a standalone component.  
- Add a Storybook entry to test various cases thoroughly.  
- Ensure proper round-trip behavior:  
  - A tester in the story should be able to create formatted text in edit mode, copy the generated string (e.g., shown in an alert callback example), paste it back to the editor content (edit or view mode), and see it rendered correctly.
- **We will only enable the formatting options that come out of the box with SlateJS**. This means the basic ones:
  - Bold, Italic, Underline, Bulleted List, Numbered List, Link, Headers, Align (left, center, right), assuming they can be enabled out of the box with SlateJS (otherwise, please do tell us to discuss together)
  - Undo and Redo if available out of the box with SlateJS
  - Any additional formatting that is easy to enable out of the box with slate JS, if available, could go in the "+" (more options)
  - We will **not** enable, even if you can see it in figma: images; emojis

This component will be reused for course content, package content, lesson notes, and anywhere that needs rich text. In particular, it will be used to render the `content` string attribute of the `AccordionItem` model, as needed in #79 .

### View mode examples

- [Accordion item](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6048-88105&t=BImsFexHPaMrk2tN-4)
Image

- [Course description](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=5731-161799&t=y24Ki2t5YiytVfpD-4)
Image

- [Package description](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6246-199514&t=BImsFexHPaMrk2tN-4)
Image

### Edit mode examples

- In the CMS
Image

- In the [CMS -- edit course](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=7274-218532&t=y24Ki2t5YiytVfpD-4)
Image

- In the [course creation pages](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=7252-300962&t=y24Ki2t5YiytVfpD-4)

Image

Image

jatinrise:
## Description

- Component name: `RichText-editor Edit and view mode` 
### This PR introduces a fully-featured Rich Text Editor and a Read-Only Renderer using Slate.js, along with Storybook integration for better UI development and testing.

- **Keyboard Shortcuts:**
- Ctrl + B → Bold
- Ctrl + I → Italic
- Ctrl + U → Underline
- Ctrl + Z/Y → Undo/Redo
   
Screenshot 2025-03-24 at 22 10 17

## Checklist

---

## Related Issues

Fixes #105 

Hey @alebg, check out my PR. I have made all the changes we discussed today.


packages/ui-kit/lib/utils/constants.tsx
```diff
@@ -0,0 +1,135 @@
...
+    Redo = "redo",
+    unlink="unlink"
+
+  }
+  
+  export const TEXT_FORMAT_OPTIONS = [
+    { id: RichTextAction.Bold, icon: <Bold  size={24}/>, label: "Bold(Ctrl+B)"},
+    { id: RichTextAction.Italics, icon: <Italic  size={24}/>, label: "Italics(Ctrl+I)"},
+    { id: RichTextAction.Underline, icon: <Underline size={24} />, label: "Underline(Ctrl+U)" },
+    {id: RichTextAction.link, icon: <Link size={24} />, label: "Link"},
```
alebg, line 50: @jatinrise please remove this Link. It makes the option appear twice:

image

But this one doesn't work (the first one, next to the Unlink, does though)

packages/ui-kit/lib/components/rich-text-editor/editor.tsx
```diff
@@ -0,0 +1,173 @@
...
+          case "y":
+            editor.redo();
+            break;
+        }
+      }
+    };
+
+    return (
+      <div className="w-full text-text-primary">
+        <Slate editor={editor} initialValue={initialValue} onChange={(value) => onChange(value)}>
```
alebg, line 154: @jatinrise as discussed yesterday, could you please serialize to a string the value that goes out of the editor?

We don't want to send out slateJS objects, but something that the backend can handle directly (strings)

alebg: It would just be something like `onChange(serialize(value))` here, assuming `serialize` is the name of your function

alebg (changes requested):
@jatinrise great job, it's looking very good. Please address my in-line comments. Apart from that:

- [x] Please change the name of the directory from `rich-text-editor` to `rich-text-element`, as it contains more than just the editor
- [ ] Please be consistent with the naming of the components. So, `RichTextEditor` is good, but please change `SlateRenderer` to its counterpart, `RichTextRenderer`. And the file name of the latter, to keep it less confusing, from `view-mode.tsx` to `renderer.tsx`. So we know quickly where to look
- [ ] In my opinion, the editor should shrink horizontally with the screen. I don't remember if there was a reason why this isn't happening?
- [ ] Let's also wait for @carlapbg 's review

packages/ui-kit/lib/components/rich-text-editor/editor.tsx
jatinrise: @alebg We can't do this in the onChange function, but we can serialize the array when clicking a button and api will call.

jatinrise:
> * RichTextRenderer

@alebg from my point of view scrollable toolbar is looking more suitable. because there are lot of function is there itovvertake the editor height

jatinrise:
@alebg, all the above inline comments have been incorporated. Can you please verify them?

carlapbg (changes requested):
Hi, in general looks so good, thank you! There is just something to change:

- Can you add the resize handle in the corner?
Screenshot 2025-03-25 at 15 01 22

- Can you move the "start typing..." text a bit lower? Right where typing begins

Screenshot 2025-03-25 at 15 08 03

- Why does the scrollbar appear at the bottom when nothing has been typed yet? When I start typing, it disappears. Can we remove it?

Screenshot 2025-03-25 at 15 11 26

- When the text is very long, the text box expands slightly to the right up to a certain point before the text starts wrapping to the next line. It would be better if this didn’t happen. The box should only grow vertically based on the text length, not to the right.

jatinrise:
@carlapbg Thank you for the review,i have done the inline comment changes,can you please verify.

carlapbg:
Thanks for the changes! Just a few things:
- The resize handle should only increase the box size in length. Right now, I can shrink it without any limits in both width and length

Screenshot 2025-03-25 at 18 08 57
Screenshot 2025-03-25 at 18 09 06


- As the screen gets smaller, the box should also shrink. Right now, on smaller screens, I have to scroll to see the full box.

Screenshot 2025-03-25 at 18 12 29

- Normally, on Mac computers, the keyboard shortcuts for editing the text (bold, underline, italic, undo/redo) use the “Command” key instead of “Control”. Could you add that? That is:  Cmd + B → Bold Cmd + I → Italic Cmd + U → Underline Cmd + Z/Y → Undo/Redo

packages/ui-kit/lib/components/rich-text-editor/editor.tsx
alebg: @jatinrise Yes I see the problem, indeed it's not that simple.

Ok so on further discussion with Mayank, let us try the following. So, let's leave the line I mentioned as is, as well as the onChange logic. Then, please pass a callback to this component, called `onLoseFocus`, which will receive a string as an input. Please implement some logic to have this callback executed once the user is not focused on the editor component anymore, and pass to this callback the serialized value.

The idea is that then, in the app (page), we will save to the backend everytime the user focuses out of the editor.

Please also adapt the stories such that you pass a console.log as this callback, so I can quickly grab the serialized values from the console and test your component more quickly.

jatinrise:
@carlapbg @alebg Thank you for the review! All the changes below have been incorporated:

- **Keyboard accessibility**: Added support for Mac users (e.g., Cmd+B, Cmd+I, etc.).
- 
- **Responsive toolbar**: The toolbar now automatically shrinks when the screen width decreases.
- 
- **Focus handling**: The onFocusLose function is now passed as a prop. Additionally, I have created a Storybook entry that returns serialized output on onBlur.
- 
- **Resizing improvements**: The editor can now only be resized vertically, and I have also fixed its minimum width.

packages/ui-kit/lib/components/rich-text-element/editor.tsx
```diff
@@ -128,45 +129,55 @@ export const RichTextEditor: React.FC<RichTextEditorProps> = React.memo(
...
             break;
         }
       }
     };
-
+   const handleBlur = () => {
+    const serialized = serialize(editor.children);
+    onLoseFocus(serialized);
+    console.log('serialized',serialized);
+   }
```
alebg, line 161: @jatinrise thanks a lot. Please don't hard-code a console.log here. What I meant is that you should be able to pass (e.g., in the story), a callback that just does console.log

jatinrise:
@alebg Thanks for the review.can you please verify now

packages/ui-kit/stories/rich-text-editor.stories.tsx
```diff
@@ -39,7 +40,8 @@ export const Basic: Story = {
     name: 'content',
     placeholder: 'Start typing...',
     initialValue: defaultInitialValue,
-    onChange: (value) => console.log('Content changed:', value),
+    onChange: (value) => console.log('Content changed:'),
```
alebg, line 43: @jatinrise please take out these onChange callbacks, they pollute the console in an inconsistent way. In the story, lets keep only the onLoseFocus callbacks that console.log, those are working nicely

alebg (changes requested):
@jatinrise please address my two inline comments. Additionally, could you please add a space between the format name and its keyboard shortcut? E.g., on hover over the "B", it should say "Bold (Ctrl+B)", instead of "Bold(Ctrl+B)"

Apart from that, really great job! It's looking very good, and your onLoseFocus implementation works nicely.

alebg:
> @alebg Thanks for the review.can you please verify now.

@jatinrise Thanks, I asked for a couple more things

packages/ui-kit/stories/rich-text-editor.stories.tsx
```diff
@@ -40,7 +40,7 @@ export const Basic: Story = {
     name: 'content',
     placeholder: 'Start typing...',
     initialValue: defaultInitialValue,
-    onChange: (value) => console.log('Content changed:'),
+    onChange: (value) => null,
```
alebg, line 43: @jatinrise Is it mandatory to pass this onChange callback always? I don't see why would we need it if we already have the onLoseFocus callback

packages/ui-kit/lib/utils/constants.tsx
```diff
@@ -44,9 +44,9 @@ import {
   }
   
   export const TEXT_FORMAT_OPTIONS = [
-    { id: RichTextAction.Bold, icon: <Bold  size={24}/>, label: "Bold(Ctrl+B)"},
-    { id: RichTextAction.Italics, icon: <Italic  size={24}/>, label: "Italics(Ctrl+I)"},
-    { id: RichTextAction.Underline, icon: <Underline size={24} />, label: "Underline(Ctrl+U)" },
+    { id: RichTextAction.Bold, icon: <Bold  size={24}/>, label: "Bold (Ctrl+B)"},
```
alebg, line 47: @jatinrise Thanks a lot! Little detail I just noticed, sorry, is that these labels should be localized as well ("Bold", "Italics", etc.)

packages/ui-kit/stories/rich-text-editor.stories.tsx
jatinrise: @carlapbg  I think this is used to solve a specific use case. Suppose we need to show both the editor and the renderer at the same time—in the assignment section for the coach, we have this use case.

packages/ui-kit/lib/utils/constants.tsx
jatinrise: @alebg Thank you for reminding me.

packages/ui-kit/lib/components/rich-text-element/types.ts
```diff
@@ -1,15 +1,16 @@
...
  * Props for the RichTextEditor component.
  */
-export interface RichTextEditorProps {
+export interface RichTextEditorProps extends isLocalAware{
   name: string;
   placeholder: string;
   initialValue: Descendant[] | undefined | string;
-  onChange: (value: Descendant[]) => void;
+  onChange?: (value: Descendant[]) => void;
   onLoseFocus?: (value:string) => void;
```
alebg, line 14: @jatinrise could you make this (`onLoseFocus`) non-optional? So when using it we know what we should pass to get the expected behaviour. If both are optional, then we can have a case in which nothing will happen from the editor outwards.

This is my last remark. After this, whenever @carlapbg validates we're ready to merge

alebg:
> @alebg Thank you! Made the onLoseFocus prop optional.

I believe you mean mandatory :). Thanks a lot, and again, great job!

carlapbg (approved):
Looks good now! Thank you!

alebg (approved):

PR #112, opened 2025-03-25
[TSK-720] [TSK-721] [TSK-722] feat(ui-kit): add HomeBanner and VideoPlayer components with storybook 

### LINKED ISSUE #75: [TSK-720] [TSK-721] [TSK-722] ui-kit: HomeBanner
Relevant model [here](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/83db9628fed062bbe34c49d1b6b8fce6aaa8d557/packages/models/src/home-page.ts#L2C1-L21C60).

For the video player, please implement a different subcomponent `VideoPlayer`, that works as a wrapper around Mux, checks for non-working video Ids, etc. It should be analogous to `UserAvatar`, and have its own story. Then, this component will use VideoPlayer (i.e., not call the Mux player directly). This can be put in the same PR as the HomeBanner, as for now video will not be shared by other PRs.

[On figma](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6241-97552&t=IrCukPhd3jcco8aU-4)

Image

manishRise11:
## Description

- Component name: `HomeBanner` 
- Figma link or screenshot: [Figma Design](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6241-97552&t=IrCukPhd3jcco8aU-4)

Image

## Checklist

---

## Related Issues

Fixes #75

### DIFF
--- a/packages/ui-kit/lib/components/video-player.tsx
+++ b/packages/ui-kit/lib/components/video-player.tsx
@@ -2,50 +2,56 @@ import MuxPlayer from '@mux/mux-player-react';
 import { useEffect, useState } from 'react';
 import { getDictionary, isLocalAware } from '@maany_shr/e-class-translations';
 
-
 export interface VideoPlayerProps extends isLocalAware {
   videoId?: string;
   thumbnailUrl?: string;
   onErrorCallback: (message: string, error: any) => void;
+  className?: string;
 }
 /**
- * A responsive, localized video player built on top of Mux, with optional lazy-loading via a thumbnail.
- * It handles error fallback, loading state, and autoplay toggling.
+ * A responsive, localized video player component built on top of Mux.
+ * 
+ * Supports optional lazy-loading via a thumbnail and handles various playback scenarios including:
+ * error fallback, loading indicator, and autoplay after interaction.
  *
- * @param videoId Mux playback ID used to load the video stream.
- * @param thumbnailUrl Optional image displayed before the video is played.
- * @param onErrorCallback A callback invoked when the video player fails to load or play.
- * @param locale Current user locale used for translations.
+ * @component
+ * @param {string} [videoId] - Mux playback ID used to load the video stream.
+ * @param {string} [thumbnailUrl] - Optional preview image displayed before the video is played.
+ * @param {(message: string, error: any) => void} onErrorCallback - Callback invoked when the video fails to load or play.
+ * @param {string} locale - Current user locale used for fetching localized text.
+ * @param {string} [className="w-full"] - Optional CSS class for customizing the outer container.
  *
- * @state showPlayer Whether to show the MuxPlayer or the thumbnail.
- * @state autoPlay Whether the video should autoplay after interaction.
- * @state videoError Whether to show the error fallback UI.
- * @state isPlayerReady Whether the Mux player has finished loading and is ready to play.
+ * @state {boolean} showPlayer - Determines whether the MuxPlayer is visible or the thumbnail is shown.
+ * @state {boolean} autoPlay - Indicates if the video should autoplay once the player is shown.
+ * @state {boolean} videoError - Indicates if the player encountered an error or no video ID was provided.
+ * @state {boolean} isPlayerReady - Tracks whether the Mux player has finished loading.
  *
  * @behavior
- * - Shows a thumbnail first (if provided), which when clicked, reveals the video player with autoplay enabled.
- * - If no thumbnail is provided, video player loads immediately.
- * - Shows a loading spinner until the Mux player is ready.
- * - Displays an error fallback UI when `videoId` is missing or the player fails to load.
- * - Uses localized text from `@maany_shr/e-class-translations`.
+ * - Displays a thumbnail initially if provided. On click, reveals the video player with autoplay enabled.
+ * - If no thumbnail is provided, loads the player immediately.
+ * - Shows a loading spinner while the player is initializing.
+ * - Displays a localized error fallback UI if the `videoId` is missing or player encounters a failure.
+ * - Uses translations from `@maany_shr/e-class-translations` for localized messages.
  *
  * @note
- * - We intentionally avoid `useCallback` here since the handler functions are not passed down or causing re-renders.
- * - If you memoize this component or lift handlers, consider wrapping callbacks with `useCallback`.
+ * - This component avoids `useCallback` as its internal handlers do not cause re-renders or prop drilling.
+ *   If memoizing the component or extracting handlers upward, consider using `useCallback`.
  *
  * @example
  * <VideoPlayer
- *   videoId="123abc"
- *   thumbnailUrl="/preview.jpg"
+ *   videoId="mux123abc"
+ *   thumbnailUrl="/assets/preview.jpg"
  *   onErrorCallback={(message, err) => console.error(message, err)}
  *   locale="en"
+ *   className="aspect-video"
  * />
  */
 export const VideoPlayer: React.FC<VideoPlayerProps> = ({
   videoId,
   thumbnailUrl,
   onErrorCallback,
-  locale
+  locale,
+  className = "w-full"
 }) => {
   const dictionary = getDictionary(locale);
   const [showPlayer, setShowPlayer] = useState(!thumbnailUrl);
@@ -70,7 +76,7 @@ export const VideoPlayer: React.FC<VideoPlayerProps> = ({
     setAutoPlay(false);
   };
 
-  const handleVideoError = (event:any) => {
+  const handleVideoError = (event: any) => {
     setVideoError(true);
     onErrorCallback(dictionary.components.videoPlayer.videoErrorText, event);
   };
@@ -79,40 +85,52 @@ export const VideoPlayer: React.FC<VideoPlayerProps> = ({
     setIsPlayerReady(true);
   };
 
+  // Fixed aspect ratio container with absolute positioning for all content
   return (
-    <div className="w-full overflow-hidden">
-      {!showPlayer && thumbnailUrl ? (
-        <img
-          src={thumbnailUrl}
-          alt="Thumbnail"
-          className="w-auto max-w-full h-auto object-contain cursor-pointer"
-          onClick={handleThumbnailClick}
-          onError={handleThumbnailError}
-        />
-      ) : videoError ? (
-        <div className="rounded-medium w-full min-w-[18rem] h-[16rem] bg-base-neutral-700 flex items-center justify-center p-4">
+    <div className={`relative ${className}`}  >
+      {/* Thumbnail state */}
+      {!showPlayer && thumbnailUrl && (
+        <div className="absolute inset-0 w-full h-full">
+          <img
+            src={thumbnailUrl}
+            alt="Thumbnail"
+            className="w-full h-full object-contain cursor-pointer"
+            onClick={handleThumbnailClick}
+            onError={handleThumbnailError}
+          />
+        </div>
+      )}
+      
+      {/* Error state */}
+      {videoError && (
+        <div className="absolute inset-0 w-full h-full bg-base-neutral-700 flex items-center justify-center p-4 rounded-medium">
           <span className="text-text-secondary text-md">
             {dictionary.components.videoPlayer.videoErrorText}
           </span>
         </div>
-      ) : (
-        <>
-          {!isPlayerReady && (
-            <div className="rounded-medium w-full min-w-[18rem] h-[16rem] bg-base-neutral-700 flex items-center justify-center p-4">
-              <div className="animate-spin rounded-full h-8 w-8 border-0   border-b-2 border-text-primary" />
-            </div>
-          )}
+      )}
+      
+      {/* Loading state - only shown when player is loading and no error */}
+      {showPlayer && !isPlayerReady && !videoError && (
+        <div className="absolute inset-0 w-full h-full bg-base-neutral-700 flex items-center justify-center p-4">
+          <div className="animate-spin rounded-full h-8 w-8 border-0 border-b-2 border-text-primary" />
+        </div>
+      )}
+      
+      {/* Video player - always rendered but hidden when not ready */}
+      {showPlayer && !videoError && (
+        <div className={`absolute inset-0 w-full h-full ${!isPlayerReady ? 'opacity-0' : 'opacity-100'}`}>
           <MuxPlayer
             key={videoId}
             streamType="on-demand"
             playbackId={videoId}
             accentColor="var(--color-base-brand-500)"
-            className={`w-full h-full ${!isPlayerReady ? 'hidden' : ''}`}
+            className="w-full h-full"
             autoPlay={autoPlay}
             onCanPlay={handlePlayerReady}
             onError={handleVideoError}
           />
-        </>
+        </div>
       )}
     </div>
   );
--- a/packages/ui-kit/lib/components/video-player.tsx
+++ b/packages/ui-kit/lib/components/video-player.tsx
@@ -10,7 +10,7 @@ export interface VideoPlayerProps extends isLocalAware {
 }
 /**
  * A responsive, localized video player component built on top of Mux.
- * 
+ *
  * Supports optional lazy-loading via a thumbnail and handles various playback scenarios including:
  * error fallback, loading indicator, and autoplay after interaction.
  *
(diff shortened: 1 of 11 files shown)


packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,63 @@
...
+            {!showPlayer && thumbnailUrl ? (
+                <img 
+                    src={thumbnailUrl}
+                    alt="Thumbnail"
+                    className="w-auto max-w-full h-auto object-contain cursor-pointer"
+                    onClick={handleThumbnailClick}
+                    onError={handleThumbnailError}
+                />
+            ) : (
+                <MuxPlayer
```
alebg, line 53: @carlapbg what should happen in the case that the video doesn't work?

In the cases in which the video doesn't work, and you click the thumbnail, the thumbnail disappears and you get a broken video forever:

- Broken video ID: video player stays in error state forever
image

- No video ID: video player stays in loading state forever
image


Is this OK, or should we do something else?

carlapbg: I think in both cases should be the "video does not exist" message. Or can we do the banner non-clickable in case the is no video?

carlapbg (changes requested):
In general it looks good. Just need to change the loading state in case there's no video or it doesn't work. Please check the inline comment I left with Luis.

packages/ui-kit/lib/components/video-player.tsx
alebg: @manishRise11 the best would be to make the banner non-clickable. If not possible, then please try to show a "Video does not exist" message in both cases

alebg (changes requested):
@manishRise11 thanks a lot. Please address the error cases for broken video IDs or missing video IDs, as discussed with @carlapbg

manishRise11:
@alebg  sir, I have handled both cases when we do not pass a videoId or a broken videoId. In both cases, I am showing a card with "Video does not exist" text with localization.

alebg (commented):
@manishRise11 thanks a lot. Code looks good to me. Let's wait for @carlapbg to validate, then we're ready to merge

carlapbg (approved):
Thank you! Looks good

alebg (approved):

alebg (approved):

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,80 @@
...
+ * <VideoPlayer
+ *   videoId="abc123xyz"
+ *   thumbnailUrl="https://example.com/thumbnail.jpg"
+ * />
+ */
+
+export const VideoPlayer: React.FC<VideoPlayerProps> = ({
+    videoId,
+    thumbnailUrl,
+    locale
```
alebg, line 29: @manishRise11 please add an onError callback here

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,80 @@
...
+        setAutoPlay(true); 
+    };
+
+    const handleThumbnailError = () => {
+        setShowPlayer(true); 
+        setAutoPlay(false); 
+    };
+
+    const handleVideoError = () => {
+        setVideoError(true);
```
alebg, line 47: @manishRise11 and call the onError callback here, so the application can inject error handling as needed. No need to remove the setVideoError here

alebg (changes requested):
@manishRise11 please check my inline comments

packages/ui-kit/lib/components/video-player.tsx
alebg: @manishRise11 Parametrization may have two things:
- A localized string with a general message, something like "Error: video not available."
- The raw error object as given by MuxPlayer's onError, if any

manishRise11:
@alebg  sir, should we pass autoPlay as props or handle it only in the component (VideoPlayer)?

jatinrise:
Heyy @carlapbg

This commit introduces a 3-second timeout to detect video loading failures and an onErrorCallback prop for better error handling.

Key Changes

- Added a timeout to trigger an error if the video takes more than 3 seconds to load.
-  Introduced onErrorCallback in props to allow external error handling.


Would it be a good idea to make the timeout duration configurable by adding loadTime as a prop instead of hardcoding 3 seconds?

carlapbg:
@jatinrise Nice! Yes, let's do the prop instead of hardcoding.

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,164 @@
+import MuxPlayer from '@mux/mux-player-react';
+import { useState, useEffect, useRef } from 'react';
+import { getDictionary, isLocalAware } from '@maany_shr/e-class-translations';
+
+export interface VideoPlayerProps extends isLocalAware {
+    videoId?: string;
+    thumbnailUrl?: string;
+    loadTimer?: number;
+    onErrorCallback?: (message: string, error: any) => void;
```
alebg, line 9: @manishRise11 I'm getting this error in the storybook:

image

I see because, in practice, the `onErrorCallback` is necessary, otherwise the component breaks. Please make the callback mandatory, and provide one (that just console logs) in the video-player story

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,164 @@
...
+ * A customizable video player that integrates with Mux for seamless video streaming.  
+ * It supports thumbnail previews, error handling, and localization.
+ *
+ * @component
+ * @param {VideoPlayerProps} props - The properties for the VideoPlayer component.
+ * @param {string} [props.videoId] - The unique identifier for the video (Mux playback ID).
+ * @param {string} [props.thumbnailUrl] - The URL of the thumbnail image displayed before playback.
+ * @param {string} props.locale - The current language/locale for text translations.
+ * @param {number} [props.loadTimer=3000] - The timeout in milliseconds before showing an error if video doesn't load.
+ * @param {(message: string, error: any) => void} props.onErrorCallback - A callback function triggered when an error occurs.
```
alebg, line 24: @manishRise11 please add here to the docstring that 'messsage' is meant to be a short error message that can be shown to the user, while `error` is more detailed and to be logged somewhere not directly to the user

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,164 @@
...
+    );
+
+    // Loading component - shown while waiting for player to initialize
+    const LoadingDisplay = () => (
+      <div className='h-full w-full items-center justify-center flex'>
+        <div className="w-6 h-6 border-4 border-base-neutral-700 border-t-transparent rounded-full animate-spin"></div>
+        </div>
+    );
+
+    const playerStyle = videoError ? { display: 'none' } : {};
```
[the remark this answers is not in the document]
alebg: @manishRise11 nevermind, just read your in-line comment. Could you please tell me what this 'flickering' is? So if you conditionally render, what happens exactly?

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,164 @@
...
+}) => {
+    const dictionary = getDictionary(locale);
+    const [showPlayer, setShowPlayer] = useState(false);
+    const [autoPlay, setAutoPlay] = useState(false);
+    const [videoError, setVideoError] = useState(false);
+    const [isLoading, setIsLoading] = useState(true);
+    const muxPlayerRef = useRef<any>(null);
+    const timeoutRef = useRef<any>(null);
+    
+    // Check if videoId exists at all
```
alebg, line 41: check

packages/ui-kit/lib/components/video-player.tsx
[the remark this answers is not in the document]
jatinrise: Hey @alebg , in my 4th case story where the video ID is incorrect, when I click on the thumbnail, it first shows the MuxPlayer and then, after a few seconds (with a flicker), it displays the error component. To handle this, I need to add a loading delay.

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,164 @@
...
+ * @param {(message: string, error: any) => void} props.onErrorCallback - A callback function triggered when an error occurs.
+ */
+export const VideoPlayer: React.FC<VideoPlayerProps> = ({
+    videoId,
+    thumbnailUrl,
+    locale,
+    loadTimer = 3000, // Default value of 3000ms (3 seconds)
+    onErrorCallback,
+}) => {
+    const dictionary = getDictionary(locale);
```
MytsV, line 33: I think it might be beneficial to utilize `useMemo` here to prevent recalculations.

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,164 @@
...
+                    onErrorCallback(dictionary.components.videoPlayer.videoErrorText, new Error('Video load timeout'));
+                }
+            }, loadTimer);
+        }
+        
+        return () => {
+            if (timeoutRef.current) {
+                clearTimeout(timeoutRef.current);
+            }
+        };
```
MytsV, line 77: I came across an issue caused by enforcing a timeout this way. Even though we display an error, the player component still remains in the tree and might still load the video. If it has an audio, it can be heard after the error component had been displayed.

I see two ways of solving this:
- Relying on timeout from the MuxPlayer component itself
- Using the reference to the MuxPlayer to stop the loading when the timeout hits

I haven't come across any flickering issues due to prolonged loading. Could you please explain in more details what bug is being observed?

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,164 @@
...
+            </span>
+        </div>
+    );
+
+    // Loading component - shown while waiting for player to initialize
+    const LoadingDisplay = () => (
+      <div className='h-full w-full items-center justify-center flex'>
+        <div className="w-6 h-6 border-4 border-base-neutral-700 border-t-transparent rounded-full animate-spin"></div>
+        </div>
+    );
```
MytsV, line 119: Moving these nested components out of the main one will prevent their recreation on each render

packages/ui-kit/lib/components/video-player.tsx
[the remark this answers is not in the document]
MytsV: > @manishRise11 could you please avoid this pattern of force-hiding by manipulating style? It's cleaner if you just conditionally render MuxPlayer: `{!videoError && !isLoading && ... }` below, like you're already doing for error and loading. In this way you don't even need the playerStyle const

@alebg this approach seems viable with the need to keep loading the stream while we need to render the loading spinner. If we render MuxPlayer conditionally, we get infinite loading as handlePlayerLoaded never gets called.

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,164 @@
...
+                <div className="relative w-full">
+                    {/* Show error display when in error state */}
+                    {videoError && <ErrorDisplay />}
+                    
+                    {/* Show loading display when still loading and not in error state */}
+                    {isLoading && !videoError && <LoadingDisplay />}
+                    
+                    {/* Always render MuxPlayer but hide it completely with CSS when in error state */}
+                    {/* This approach allows error events to be captured without visible flickering */}
+                    <div style={playerStyle} className={isLoading && !videoError ? "opacity-0" : "opacity-100"}>
```
MytsV, line 147: The approach of using opacity to hide the MuxPlayer causes the loading spinner to not be centered. There's also a small visual bug of it appearing on top of the player for a few milliseconds.

Replacing this with `"hidden" : "block"` for display property seems to fix both these issues

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,164 @@
...
+                    src={thumbnailUrl}
+                    alt="Video Thumbnail"
+                    className="w-auto max-w-full h-auto object-contain cursor-pointer"
+                    onClick={handleThumbnailClick}
+                    onError={() => {
+                        if (!videoError) {
+                            setShowPlayer(true);
+                        }
+                    }}
+                />
```
MytsV, line 136: A minor thing a noticed is thumbnail's alt text being show for a fraction of a second if there's an error, similar to the flicker case with the MuxPlayer described above. Could this be handled similarly with display property relying on something like isImageLoading? Then this thumbnail component could be extracted outside VideoPlayer for better readability.

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,164 @@
...
+    // Check if videoId exists at all
+    useEffect(() => {
+        if (!videoId || videoId.trim() === '') {
+            setVideoError(true);
+            setIsLoading(false);
+            onErrorCallback(dictionary.components.videoPlayer.videoErrorText, new Error('Missing video ID'));
+        } else {
+            setVideoError(false);
+            // Only show player if there's a valid ID and no thumbnail, or if user clicked thumbnail
+            setShowPlayer(!thumbnailUrl);
```
MytsV, line 50: Should this be set if we check for thumbnailUrl being undefined below on render?

packages/ui-kit/stories/hero.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+export const ShortDescription: Story = {
+  args: {
+    title: 'Quick Start Guide',
+    description: 'Get started in minutes!',
+    videoId: 'YEK2woOv35g5CTTcbxtiNjG3kRH9ddCChjxATEpVTxc',
+    thumbnailUrl:
+      'https://res.cloudinary.com/dryynqhao/image/upload/v1742541099/lrpuzzgdayhoirs4gqgj.png',
+      locale: 'en',
+  },
+};
```
MytsV, line 61: For this story I see the layout kind of shifting after loading the video and feel like that should be addressed.

image
image

packages/ui-kit/lib/components/video-player.tsx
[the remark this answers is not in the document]
alebg: > @alebg this approach seems viable with the need to keep loading the stream while we need to render the loading spinner. If we render MuxPlayer conditionally, we get infinite loading as handlePlayerLoaded never gets called.

@MytsV Sorry, i didn't understand. What is the suggestion for improvement here then? Do we or do we not use conditional rendering?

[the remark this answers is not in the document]
MytsV: > > @alebg this approach seems viable with the need to keep loading the stream while we need to render the loading spinner. If we render MuxPlayer conditionally, we get infinite loading as handlePlayerLoaded never gets called.
> 
> @MytsV Sorry, i didn't understand. What is the suggestion for improvement here then? Do we or do we not use conditional rendering?

We should leave the component in the tree, so no conditional rendering. There are suggestions below in the review on how to prevent some visual bugs like the loading spinner not being centered

jatinrise:
Hey @alebg,
I've implemented an alternative approach that I believe is more optimized. I haven't used useCallback in this case, as the handler functions aren't being passed down to child components or causing unnecessary re-renders.

packages/ui-kit/lib/components/video-player.tsx
```diff
@@ -0,0 +1,119 @@
...
+  onErrorCallback,
+  locale
+}) => {
+  const dictionary = getDictionary(locale);
+  const [showPlayer, setShowPlayer] = useState(!thumbnailUrl);
+  const [autoPlay, setAutoPlay] = useState(false);
+  const [videoError, setVideoError] = useState(!videoId);
+  const [isPlayerReady, setIsPlayerReady] = useState(false);
+
+  useEffect(() => {
```
alebg, line 56: @MytsV could you please re-review this? it was re-implemented

MytsV: I like the new implementation, but I still experience bugs with thumbnail flickering and layout shifting described above and feel like they should be addressed.

alebg: @jatinrise could you please address @MytsV 's comment?

MytsV:
@jatinrise looks great! I just noticed a minor visual inconsistency with the error view having rounded borders in contrast to the loading view. Pushed the change myself as I will need this component for complete page implementation.

alebg (approved):

PR #116, opened 2025-03-26
feat(dropdown): add position prop for customizable dropdown placement

ObitoDevv:
The footer dropdown was opening `downward` instead of upward as intended, so the Dropdown component has been updated to correctly handle the `position='top'` prop by applying `bottom-full mb-2 ` classes, while preserving the default `position='bottom' ` behavior for other uses like the NavBar.

### DIFF
--- a/packages/ui-kit/lib/components/dropdown.tsx
+++ b/packages/ui-kit/lib/components/dropdown.tsx
@@ -22,6 +22,55 @@ export interface DropdownProps {
   position?: 'top' | 'bottom';
 }
 
+/**
+ * A reusable Dropdown component supporting single-select and multi-select variants with truncation and hover tooltips.
+ * Variants include a simple dropdown, a radio-button-based color selector, and a searchable multi-select with checkboxes.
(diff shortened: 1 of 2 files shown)


alebg (commented):
Looks good to me @ObitoDevv . Let's wait for @carlapbg 's review and we're ready to merge

carlapbg (changes requested):
Hi! Just one thing: Can you add a hover effect in case the text is too long? When the text is too long, the chevron icon disappears. 

Screenshot 2025-03-26 at 16 15 50

Also, the box should not grow; it should remain the same size even if the text is long. 

Screenshot 2025-03-26 at 16 21 18

ObitoDevv:
@carlapbg  Hi mam, have fixed the issue. Can you please verify it ?

ObitoDevv:
@carlapbg Fixed the issue for multi-select-options Mam.

carlapbg (approved):
Looks good now, thank you!

alebg (approved):

PR #117, opened 2025-03-26
[TSK-668] Add a production Dockerfile and supporting configuration

### LINKED ISSUE #66: [TSK-668] Create a Dockerfile for apps/platform

MytsV:
## Description

This PR adds configuration needed to dockerize the platform application for production. Multi-stage build isn't utilized due to the need of debugging the application in production.

The Apache configuration for the proxy server was edited, as there were issues integrating it with the platform application due to (null) values getting into X-Forwarded headers.

The pm2 is now run with --deamon, as otherwise the shell of the docker container gets blocked.

The issue with eslint plugin has creeped into the PM2 environment for an unknown reason. The Node.js version used in the environment is confirmed to be the latest (23.10.0), and running any NX command outside of PM2 doesn't lead to this error. As a temporary fix, nx.json is processed during image build to not contain the failing plugin as it's not relevant to production environments.

## Checklist

- [x] Ran the docker container from the built image without SSL
- [ ] Ran the docker container with SSL configured

## Related Issues

Fixes #66


config/README.md
```diff
@@ -0,0 +1,138 @@
...
+    -e NEXT_PUBLIC_E_CLASS_PLATFORM_LOGO_URL=<https://your-logo-url.com/logo.png> \
+    -e NEXT_PUBLIC_E_CLASS_PLATFORM_SHORT_NAME=<dev> \
+    -p 80:80 -p 443:443 \
+    --mount type=bind,source=your/hostcert.pem,target=/etc/dad/hostcert.pem \
+    --mount type=bind,source=your/hostkey.pem,target=/etc/dad/hostkey.pem \
+    --mount type=bind,source=your/ca.pem,target=/etc/dad/ca.pem \
+    maany/e-class-platform
+```
+
+```bash
```
alebg, line 39: @MytsV what is this other command for? Not asking to remove it, but maybe just explain it please? since it's a bit different

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN pnpm install --frozen-lockfile
+
+# This removes a plugin that is not required in production and causes the build to fail.
+# The build failure happens only in PM2 environment despite utilizing the latest Node version.
+RUN jq '.plugins = [.plugins[] | select(.plugin != "@nx/eslint/plugin")]' nx.json > nx.tmp.json && \
+    mv nx.tmp.json nx.json
+
+EXPOSE 443
+EXPOSE 80
+EXPOSE 3000
```
alebg, line 44: @MytsV I don't see port 3000 exposed in websat, just to understand, why do you decided to add it here?

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.2/install.sh | bash \
+    && . $NVM_DIR/nvm.sh \
+    && nvm install $NODE_VERSION \
+    && nvm alias default $NODE_VERSION \
+    && nvm use default
+
+RUN npm i -g pm2 pnpm
+
+WORKDIR /opt/dadai/app
+COPY . .
```
alebg, line 31: @MytsV just wondering here, do we need to bundle the whole monorepo? i.e., as opposed to just bundle apps/platform

config/README.md
MytsV: I copied it from websat, and it seems to represent running the HTTPS server without logging. I agree it might be redundant, so I will remove it.

Dockerfile
MytsV: Per discussion with @maany, there is a need for debugging the application in development mode. For that, using a multi-stage build won't work, as the apps/platform depends on other components of the monorepo.

Dockerfile
MytsV: So the EXPOSE command in Dockerfile works, as I understand, mostly as a hint. Usually ports are forwarded explicitly with -p, and those withing EXPOSE statements are attached to random host ports only with the -P option. Here for debugging purposes it might be useful to expose the application behind the Apache server, as I had it fail for me due to some incorrectly set headers. The current mechanism of authentication will be extra sensitive to those issues.

alebg (changes requested):
@MytsV alright, looks good! Just one thing we just discussed with Mayank, could you please put all of these files in `apps/platform` instead of on the root of the monorepo? The reason is that they conceptually belong to platform, not to the whole project.

It should then be possible to build this properly with the whole monorepo inside with something like `docker build -f apps/platform/Dockerfile . -t <tag>` (from the root folder), or something like that. Maybe `docker build -f Dockerfile ../.. -t <tag>`, if you're already in `apps/platform`. Please give it a test

MytsV:
@alebg thank you for the review! I have updated the configuration to exist in apps/platform folder and ensured that the build works with provided instructions. The README.md now contains them as well.

alebg (approved):
@MytsV perfct, thanks a lot

PR #124, opened 2025-03-29
[TSK-857] Import e-class-core models and update dependencies

alebg:
Integrate the `@maany_shr/e-class-core`  to start using models from there. Test with a single ui-kit file using @maany_shr/e-class-core/models

## TODO

This will fail `platform:build`. 

- [ ] Find a way of making TS understand correctly typing from @maany_shr/e-class-core in `platform:build`. Note that `ui-kit:build` works properly

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
   },
   "private": true,
   "dependencies": {
+    "@maany_shr/e-class-core": "^1.7.5",
     "@tailwindcss/postcss": "4.0.0-beta.8",
     "i18n": "^0.15.1",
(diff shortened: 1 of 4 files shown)


alebg:
@maany couldn't make this work. Tried tweaking different configs but with no luck. The closest I got to a (at least ugly) fix was changing `    "moduleResolution": "node",` to `    "moduleResolution": "bundler",` in `tsconfig.base.json`. However this created new type errors on the use of the profile model. Annoyingly, these type errors were *not* reported when changing back to use the local `models` package, that contains the exact same `profile` model.

PR #128, opened 2025-04-02
[TSK-850] Refactor components to take out props from State, and properly wire state changes via callbacks, Part 1

### LINKED ISSUE #125: [TSK-850] Refactor components to take out props from State, and properly wire state changes via callbacks, Part 1
In general, the use of React hooks should be kept to a minimum, only used when strictly necessary. This is to prevent performance hits and unexpected behaviour when running the app. We've overlooked certain patterns found in code in the main branch, that have risen up upon further analysis / discussion. 

## Overview of detected problems in certain merged components

- (a) Storing props in State disconnects the data of the prop from its original source, making it dependent only on the ui-kit component itself. E.g., the app loses control over this. More concretely, this implies that later changes (e.g., because the parent passes a new prop), `useState` (or any other State hook) will not update the state automatically

A simple solution is to simply not use React hooks unless necessary. For example, the rich text editor needs to manage its own state after being passed an initialValue. This is a case in which we can put initialValue in a State and hand it over the component: we don't need to sync-up initialValue anymore after the initial render.

But only few cases are like this. In many cases, we do need to keep the prop as passed by the parent in sync with the ui-kit component. Disconnecting this by storing props in State may cause unexpected behaviour and bugs throughout the application.

- (b) ui-kit components should be kept 'dumb', and almost all functionality should be passed from the outside as callbacks. The ui-kit shouldn't set what the application must do: it should be the other way around.

## Cases to refactor

We'll focus on anything regarding the Home Page first, then do a part 2 of refactoring if necessary.

- `NavBar` [manages the locale State independently](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/b021957c0c1218f8b022bccdc408a1db8e3cb10a/packages/ui-kit/lib/components/navbar.tsx#L63). This is unnecessary, as the locale should just be passed from the parent as a prop, and then the ui component should just execute the onChangeLanguage callback when the user changes it. No need to use State at all here.

- Something similar happens in the open PRs for the [`Footer`](https://github.com/dream-aim-deliver/dad-e-class-monorepo/pull/123/files#diff-5e155b90fd4810145ebed8e80dea84d6119967809b3b3a4fb23e07e85d7ff603R52), and [`RecentActivity`](https://github.com/dream-aim-deliver/dad-e-class-monorepo/pull/123/files#diff-5e155b90fd4810145ebed8e80dea84d6119967809b3b3a4fb23e07e85d7ff603R52)

- `Dropdown` (used in the NavBar) [stores `options` in useState](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/b021957c0c1218f8b022bccdc408a1db8e3cb10a/packages/ui-kit/lib/components/dropdown.tsx#L100-L106)

- `GeneralCard` [instantiates its button callback within itself](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/b021957c0c1218f8b022bccdc408a1db8e3cb10a/packages/ui-kit/lib/components/carousel/generalcard.tsx#L65).  This is not good since the UI kit is mandating what the functionality of the button is. This prevents us from passing from the app more appropriate functionality, e.g., Next.js routing. In this case, the callback should just be passed from the outside

## Tasks

- [x] NavBar, fixed in #128 
- [x] Footer, fixed in #122 
- [x] RecentActivity, fixed in #123 
- [x] Dropdown, fixed in #128 
- [x] General Card, fixed in #128

ObitoDevv:
## Description

- Component name: `navbar-refactored ` , `General Card-refactored` , and `Dropdown-refactored`

---

## Checklist

---

## Related Issues

Fixes #125

### DIFF
--- a/packages/ui-kit/lib/components/carousel/generalcard.tsx
+++ b/packages/ui-kit/lib/components/carousel/generalcard.tsx
@@ -4,19 +4,22 @@ import { Badge } from "../badge";
 import { getDictionary, isLocalAware } from "@maany_shr/e-class-translations";
 import { homePage } from "@maany_shr/e-class-models";
 
-interface GeneralCardProps extends homePage.TGeneralCard, isLocalAware {}
+interface GeneralCardProps extends homePage.TGeneralCard, isLocalAware {
+  onButtonClick: () => void;
+}
 
 /**
  * A card component for displaying general information, such as course or content details.
- * Includes an image, title, description, optional badge, and a button with a link.
+ * Includes an image, title, description, optional badge, and a button with a custom callback.
  *
  * @param imageUrl The URL of the card's image. Displays a placeholder if omitted or invalid.
  * @param title The title of the card content.
  * @param description A brief description of the card content.
  * @param badge Optional text for a badge displayed over the image (e.g., "New", "Featured").
  * @param buttonText The text displayed on the button.
- * @param buttonUrl The URL to navigate to when the button is clicked.
+ * @param buttonUrl The URL associated with the button (can be used in the callback).
  * @param locale The locale for translation and localization purposes (e.g., "en" for English, "de" for German).
+ * @param onButtonClick Callback function to execute when the button is clicked.
  *
  * @example
  * <GeneralCard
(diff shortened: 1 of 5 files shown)


carlapbg (approved):
Looks good in the design! Thank you

alebg (approved):
looks good, thanks a lot @ObitoDevv !

PR #133, opened 2025-04-07
feat(input-field): add inputClassName prop for custom input styling

### LINKED ISSUE #132: fix(ui-kit): add inputClassName prop to TextField component to allow input customization
Currently, the TextField component in ui-kit does not accept a custom class for the underlying <input> element.

jatinrise:
Added ` inputClassName `prop to TextField component to allow input customization and also done changes in component where we have used the text field component.

fix #132

### DIFF
--- a/packages/ui-kit/lib/components/buy-coaching-session.tsx
+++ b/packages/ui-kit/lib/components/buy-coaching-session.tsx
@@ -110,7 +110,8 @@ function BuyCoachingSession({ courses, onClick, locale,currencyType }: BuyCoachi
                             </button>
                             <InputField
                                 type='number'
-                                className='w-[3rem] h-[3rem] text-lg text-center'
+                                className='w-[3rem] h-[3rem] text-lg'
+                                inputClassName='text-center'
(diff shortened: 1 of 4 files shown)


jatinrise:
Heyy @alebg sir, please review this pr at first place.

alebg (commented):
@jatinrise could you please give me a bit of context?

1. Why is this necessary?
2. Why are you refactoring the profile component?

jatinrise:
Hey @alebg, sure!

It’s necessary to add the` inputClassName` prop to the custom <input> in InputTextField so we can control the input text alignment depending on the context where it's used.

Previously, in the `Profile componen`t, the input wasn't taking the full width by default and the text was aligned to the center — which wasn’t ideal. To fix this and ensure consistent behavior across components, I made it customizable.
For instance, in the `BuySession component`, we want the text to be centered, but elsewhere (like Profile), it should remain left-aligned by default.

Let me know if anything needs more context!

alebg (approved):

PR #143, opened 2025-04-14
[TSK-866] Add exceptions to .editorconfig

MytsV:
This PR adds differing rules for some file formats in .editorconfig. This will facilitate the development process by setting a correct indent for Python or TypeScript files.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,13 +1,32 @@
 # Editor configuration, see http://editorconfig.org
 root = true
 
+# Default settings for all files
 [*]
 charset = utf-8
+end_of_line = lf
 indent_style = space
 indent_size = 2
 insert_final_newline = true
 trim_trailing_whitespace = true
 
+# Python files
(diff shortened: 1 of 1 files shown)


alebg (commented):
Thanks @MytsV . Just one question, why do we need a 'Makefile' config?

Apart from that, just as a remark, this will not affect VSCode users unless we download an extension.

MytsV:
> Thanks @MytsV . Just one question, why do we need a 'Makefile' config?
> 
> Apart from that, just as a remark, this will not affect VSCode users unless we download an extension.

Thank you for the review! The Makefile was something coming from the AI generated config. I left it as something that really stood out. But I agree that it might be redundant as we're not planning on using those. Pushed the commit without the Makefile to both repositories.

The .editorconfig just really affects my work in JetBrains IDE, and I feel like this might be beneficial for contributors with this setup.

alebg (approved):

PR #144, opened 2025-04-14
[TSK-920] Include base-brand color in @theme block

### LINKED ISSUE #142: [TSK-920] Fix Tailwind config issues
As noted by @MytsV during the preparation of the HomePage for RM#4 end-meeting demo. We'll provide more details here soon.

MytsV:
This PR ensures the brand palette can be used in components without defining intermediate variables. This might be useful for preparing quick demo components in the future.

The data-theme attribute approach doesn't have any benefits apart from being a convention. As this change will require resolving conflicts with some ongoing work, it seems better to leave it as is.

Edit by LB: Fix #142

### DIFF
--- a/packages/ui-kit/lib/assets/css/tailwind.css
+++ b/packages/ui-kit/lib/assets/css/tailwind.css
@@ -60,6 +60,17 @@
   --color-base-neutral-800: var(--color-base-neutral-800);
   --color-base-neutral-900: var(--color-base-neutral-900);
   --color-base-neutral-950: var(--color-base-neutral-950);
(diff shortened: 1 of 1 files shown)


MytsV:
> > The data-theme attribute approach doesn't have any benefits apart from being a convention. As this change will require resolving conflicts with some ongoing work, it seems better to leave it as is.
> 
> Hey @MytsV , when you say "this change", you're referring to the hypothetical change to a data-theme attribute approach, right? You're not referring to the actual changes in this PR?

Yes, the change in this PR is unbreaking :)

alebg (approved):

PR #147, opened 2025-04-17
[TSK-710] [TSK-711] [TSK-712] migration(ui-kit): ReviewCoachingSessionModal

### LINKED ISSUE #72: [TSK-710] [TSK-711] [TSK-712] migration(ui-kit): ReviewCoachingSessionModal
[[On figma]](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=8453-299929&t=zyhReeZZP4aZpx5y-4)

Image

ObitoDevv:
## Description

- Component name: `ReviewCoachingSessionsModal` 
- Figma link or screenshot: (https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=8453-299929).

Image

---

## Checklist

---

## Related Issues

Fixes #72


packages/ui-kit/lib/components/review-coachingsession-modal.tsx
```diff
@@ -0,0 +1,167 @@
...
+        setSubmitted(true);
+    };
+
+    const handleClose = () => {
+        if (onClose) {
+            onClose();
+        }
+    };
+
+    const handleSkip = () => {
```
alebg, line 40: @ObitoDevv I think it may be a good idea to pass to the outside something like a boolean, "skipped" that is True if the process was skipped, and False if the user filled the form and submitted. In this way, we can ignore the first case cleanly

packages/ui-kit/lib/components/review-coachingsession-modal.tsx
```diff
@@ -0,0 +1,167 @@
...
+        }
+        handleClose();
+    };
+
+    const isFormValid = rating > 0 && review.trim().length > 0;
+    const dictionary = getDictionary(locale);
+
+    if (!submitted) {
+        return (
+            <form
```
alebg, line 52: @ObitoDevv it would be a good idea to add some load state + error handling. E.g., imagine the user clicks the submit button, but the backend fails. We wouldn't want users to see the success part of the modal if that happens, and then they would just simply lose the information put in the review. If you emit some sort of error instead, and prevent the user from getting to the success view, then the page / parent component could show a toast or something, and eventually tell the user to retry later but to store his review elsewhere, etc.

ObitoDevv:
@alebg Thanks for the feedback! Sir , I've incorporated both suggestions: 
1. Added the `skipped` boolean to `onSubmit` (false) and `onSkip` (true) to clearly distinguish submitted vs. skipped reviews.
2. Implemented loading state with `isLoading` and error handling to prevent the success view on backend failure, preserving user input with an error message.

Regarding the SVG circle spinner in the loading state, I noticed it's not in the Figma designs. Should we stick with this implementation, or is there a specific spinner design we should use instead? Let me know!

packages/ui-kit/lib/components/review-coachingsession-modal.tsx
```diff
@@ -0,0 +1,242 @@
...
+    const [submitted, setSubmitted] = useState(false);
+    const [neededMoreTime, setNeededMoreTime] = useState(false);
+    const [isLoading, setIsLoading] = useState(false);
+    const [error, setError] = useState<string | null>(null);
+
+    const handleReviewSubmit = async (e: React.FormEvent) => {
+        e.preventDefault();
+        if (!onSubmit) return;
+
+        setIsLoading(true);
```
alebg, line 63: @ObitoDevv So here we prefer the approach of passing `isLoading` and `isError` as booleans as props. There are a couple of reasons for this. Basic one is that we want to keep the ui-kit as dumb as possible, and just decide the app itself to check when it wants to trigger loading, error, success states, etc. Sorry if this wasn't clear enough, but this criterion also applies here.

So here, in your implementation, your component is deciding the error state by a try-catch, whereas it should be the responsibility of the app to decide what an error is, not the ui-kit.

Another reason, in the error case, is that if the error is handled in a useState like you do here, it will disconnect the state from the rest of the app. If I change the props by selecting another locale, for example, everything will change to German as expected, except your error text that will still render in English:

immagine

The approach of passing isLoading and isError from the outside is just plain cleaner, spares us from using React hooks, and makes it easier for you to implement stories, since now all you have to do is expose a boolean toggle for each.

alebg (changes requested):
@ObitoDevv thanks a lot! Please address my inline comment on how to implement loading and error states via props.

Apart from that, one thing I noticed right away, is that if I fill some data but then click 'Skip', your component will anyway render the success case, which is not good UX:

immagine

immagine

About the spinner, I will come back to you, we may need to implement it as its own component.

packages/ui-kit/lib/components/review-coachingsession-modal.tsx
```diff
@@ -0,0 +1,242 @@
...
+    if (!submitted) {
+        return (
+            <div className="flex flex-col items-end gap-4 p-6 rounded-lg border border-card-stroke bg-card-fill max-w-[340px] shadow-[0_4px_12px_rgba(12,10,9,1)]">
+                <form
+                    onSubmit={handleReviewSubmit}
+                    className="flex flex-col items-end gap-4 w-full"
+                >
+                    <label
+                        htmlFor="courseRating"
+                        className="text-lg font-bold leading-tight text-white text-justify w-full"
```
carlapbg, line 101: Can you delete the 'text-justify' or change it to 'text-left' so there aren’t many gaps between the words please
Screenshot 2025-04-20 at 21 35 14

carlapbg (changes requested):
Hi! I have one inline comment. Thanks!

ObitoDevv:
Hi @aleg sir, Thanks a lot for the review.

This is the updated ReviewCoachingSessionModal:

1. Skip button now always enabled (except during loading).
2. Error message sourced from dictionary.components.reviewCoachingSessionModal.errorState.
3. Used IconLoaderSpinner for loading state.
4. Send Review button enabled when rating is provided.

Just one question is remanning - what to show in the tooltip description? 

Please review and let me know if any further changes are needed!

alebg:
@carlapbg what should be put in the tooltip text?

alebg (changes requested):
@ObitoDevv thanks, this is looking better. Could you please put back in the story the alert callbacks? They were good to see, in the success case, that the data was being passed outside.

Also, just a question, if we use the error case, how do we prevent the data that the user inputted from being deleted? Could you give it a try and compose a story in which the callback on submit sets the error state? so that we can be sure that the data is preserved

carlapbg:
Hi! If I write a review and then click “skip,” it still sends the review. It should just close the review without sending it.

And about the tooltip description, I don’t think we need it in this case. We can just have the title “Your review” and the placeholder in the text area, that is clear enough.

ObitoDevv:
Hi @aleg sir,

I've raised PRs for the following updates:

1. Coaching Session Component: Created a new branch to resolve conflicts with the main branch and pushed the complete code using the Children Approach.

2. CourseIntroBanner Component: Added a new CourseIntroBanner component.

3.SideMenu Component: Updated height from h-full to h-fit for consistent height in collapsed and expanded states, as suggested by Carla mam.

4.ReviewCoachingSessionModal: Fixed the issue where the success view was shown instead of the error state in the ErrorStateWithPreservedData story. Updates include:

- Added submitted and setSubmitted props to control success view.
- Removed automatic setSubmitted(true) in handleReviewSubmit.
- Used isError to preserve user inputs in error state.
- Updated Storybook story to simulate failed submission with a 1-second delay, keeping submitted as false.

Regarding CourseCardList and CourseContainer, @ObitoDevv confirmed no changes are needed. The empty state width issue in Storybook is due to missing parent context and will work correctly in the actual page.

carlapbg (dismissed):
Thanks! Looks good to me

packages/ui-kit/lib/components/review-coachingsession-modal.tsx
```diff
@@ -0,0 +1,217 @@
...
+import { IconButton } from './icon-button';
+import { IconClose } from './icons/icon-close';
+import { IconLoaderSpinner } from './icons/icon-loader-spinner';
+import { getDictionary, isLocalAware } from '@maany_shr/e-class-translations';
+import { TextAreaInput } from './text-areaInput';
+
+export interface ReviewFlowProps extends isLocalAware {
+  onClose?: () => void;
+  onSubmit?: (rating: number, review: string, neededMoreTime: boolean, skipped: boolean) => void;
+  onSkip?: (skipped: boolean) => void;
```
alebg, line 16: @ObitoDevv why do we need this boolean?

packages/ui-kit/lib/components/review-coachingsession-modal.tsx
```diff
@@ -0,0 +1,217 @@
...
+    onSubmit(rating, review, neededMoreTime, false);
+  };
+
+  const handleClose = () => {
+    if (onClose) {
+      onClose();
+    }
+  };
+
+  const handleSkip = () => {
```
alebg, line 82: @ObitoDevv as discussed many times over, this kind of logic shouldn't be in the ui-kit. All of this should be controlled by the parent via the onSkip callback

alebg: @ObitoDevv Also the onClose and onSkip should be independent from each other, just in case we want fine control

packages/ui-kit/lib/components/review-coachingsession-modal.tsx
```diff
@@ -0,0 +1,217 @@
...
+    if (onSkip) {
+      onSkip(true);
+    }
+    // Reset form state
```
alebg, line 86: @ObitoDevv i.e., you can just remove lines 86 to 94

packages/ui-kit/lib/components/review-coachingsession-modal.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+  const dictionary = getDictionary(locale);
+  const isFormValid = rating > 0;
+
+  const handleReviewSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+    if (!onSubmit || !isFormValid) return;
+
+    // Call onSubmit and let parent/story handle submitted state
+    onSubmit(rating, review, neededMoreTime, false);
```
alebg, line 73: onSubmit doesn't need the last boolean. rating, review, and neededMoreTime are enoguh

alebg (changes requested):
@ObitoDevv thanks, please address my inline comments

alebg:
@ObitoDevv also, the link to figma is incorrect. Next time please just copy it from the issue. Here's the correct one: https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=5731-181090&t=8GmdhPKbh3tkKh6f-0

ObitoDevv:
@alebg Sir,  Thank you for your feedback. I've made the following changes:

1. Removed the skipped boolean parameter from both onSubmit and onSkip interfaces as it was unnecessary
2. Simplified onSkip to just be a callback without parameters
3. Removed form reset logic from the component (lines 86-94) to follow the principle of keeping logic in the parent component
4. Made onClose and onSkip completely independent from each other
5. Updated the component's JSDoc comments to reflect these interface changes
6. Also updated the Storybook file to align with the new component interface

alebg (approved):

PR #150, opened 2025-04-19
[TSK-776] [TSK-777] [TSK-778] ui-kit: CoachCardList

### LINKED ISSUE #93: [TSK-776] [TSK-777] [TSK-778] ui-kit: CoachCardList
[In Figma](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6340-45343&t=pUvzvGlbr8ychX2x-4).

Image

EDIT: Should receive as part of its props ~`coachCards: CoachCard[];`~  `children` (meant to be the cards), as well as a title

jatinrise:
## Description

In this PR, I have added a  `CardListLayout `component to showcase a responsive card layout. Inside this layout component, I have created the `CoachCardList` component that renders individual CoachCard components. The CoachCardList component is fully responsive, adjusting automatically based on the screen size.

On desktop screens, the layout displays 6 cards in a row.

On tablet screens, it shows 4 cards.

On mobile screens, only 2 cards are shown per row.

- Component name: `CardListLayout` ,`CoachCardList`,
- Figma link or screenshot: 
We need to handle a few edge cases in the card layout:

1. When all coaches have the same amount of data (e.g., the same number of skill badges):

Screenshot 2025-04-19 at 14 14 45

2.When the data varies between coaches (e.g., some have more badges while others have fewer):

Screenshot 2025-04-19 at 14 28 43

---

## Checklist

- [ ] My changes align with the **core models** that have been discussed and agreed upon.

---

## Related Issues

Fixes #93


packages/ui-kit/stories/coach-card-list.stories.tsx
```diff
@@ -0,0 +1,193 @@
+import React from 'react';
+import { Meta, StoryObj } from '@storybook/react';
+import CoachList from '../lib/components/coach/coach-card-list';
+import { CoachCardDetails } from '../lib/components/coach/coach-card';
+
+const meta: Meta<typeof CoachList> = {
```
alebg, line 6: @jatinrise small detail, please call it `CoachCardList` to keep it consistent

packages/ui-kit/lib/components/coach/coach-card-list.tsx
```diff
@@ -0,0 +1,56 @@
...
+    title?: string;
+    coaches: CoachCardDetails[];
+    onClickBookSession?: () => void;
+    onClickViewProfile?: () => void;
+}
+
+export default function CoachList({ coaches,title,onClickBookSession,onClickViewProfile,locale }: CoachListProps) {
+  const [visibleCards, setVisibleCards] = useState(6);
+  const allCoaches = [...coaches];
+  useEffect(() => {
```
alebg, line 24: @jatinrise So, we had an internal discussion on this UX-wise, and we've decided to override Stefano's design: we will not remove cards when the screen shrinks, so you can always see all the relevant information and offers. 

Please just always show all cards passed as children. Just have them stack. And keep the amount of cards at 3 per row on desktop, 2 per row on tablet, 1 per row on mobile (as it is now already).

alebg: @jatinrise this should be doable with `flex` and so on, I don't think we need react hooks in this component at all.

packages/ui-kit/lib/components/coach/coach-card-list.tsx
```diff
@@ -0,0 +1,56 @@
...
+    window.addEventListener("resize", handleResize);
+    return () => window.removeEventListener("resize", handleResize);
+  }, []);
+  const displayedCoaches = allCoaches.slice(0, visibleCards);
+  
+  return (
+    <div className="flex flex-col gap-10">
+    <h3 className="text-text-primary lg:text-[40px] text-2xl">{title}</h3>
+    <CardListLayout>
+      {displayedCoaches.map(coach => (
```
alebg, line 45: @jatinrise please use the children as props strategy here, and if possible remove all react props so this can be a server side component

alebg (changes requested):
@jatinrise please address my in-line comments. Overall I think we don't need two list components here. Since we're not following Stefano's figma on the card removal anymore, I think you should just need the card list layout (server side component, takes children as props) and the coach card. Then in the story you can assemble an example coach-card-list, passing coach cards as children to the card list layout

Does this sound good?

alebg:
@carlapbg When you start shrinking the screen, at some point the cards overlap:

immagine

To me this doesn't look good, what do you think?

jatinrise:
> @jatinrise As a follow-up to my in-line comments, overall I think we don't need two list components here. Since we're not following Stefano's figma on the card removal anymore, I think you should just need the card list layout (server side component, takes children as props) and the coach card. Then in the story you can assemble an example coach-card-list, passing coach cards as children to the card list layout
> 
> Does this sound good?

Thank you @alebg, got it!.

jatinrise:
> @carlapbg When you start shrinking the screen, at some point the cards overlap:
> 
> immagine
> 
> To me this doesn't look good, what do you think?

Hey @alebg , I’ll remove the logic for the visible card—then it should look fine.

packages/ui-kit/lib/components/coach/coach-card-list.tsx
jatinrise: "Hey @alebg, I was using the `CardListLayout` component where I'm passing the list of cards as children. If I remove` displayedCoaches` and `title `as props, how can we pass the data to the `CoachCardList `component? And if you're talking about server-side rendering, it should work fine since we're not using any client-side hooks."

carlapbg (changes requested):
Hi! Just some things:

- There ir a huge gap between the skills and the description. Can you fixed it?

Screenshot 2025-04-20 at 22 17 26

- Apart from that, I still get that problem when shrinking than Luis mentioned. The gap between each card should be always the same.
Screenshot 2025-04-20 at 21 40 40

- The mobile view cuts and to see it full we need to scroll to the right. Can you make the component more responsive so it will get smaller in mobile devices?
Screenshot 2025-04-20 at 22 26 26

jatinrise:
Thank you, @alebg and @carlapbg. All the inline changes have been incorporated.

alebg (dismissed):
@jatinrise thanks a lot! looks very good now! I'll wait for @carlapbg for a final check and then merge this if everything is good. Please note that I did a minor cleanup and pushed a commit, including wiring properly the story callbacks to test that they work when you click the cards' buttons

carlapbg:
@jatinrise Thanks! Just a few observations: 

I see that the gap is still there. I think the best option in this case is to have only 2 lines in the skills section instead of 3 like it is now. That way, if a coach has only a few skills, the gap won’t look so big.
Screenshot 2025-04-22 at 13 42 57


Also, in the same skills section, when a coach has several and the "+4 more..." appears, for example, only 3 additional skills actually show up. There's always one skill less than what the button indicates. Please make sure this gets corrected so it displays the actual number of skills.

Screenshot 2025-04-22 at 14 20 52
Screenshot 2025-04-22 at 14 21 21

jatinrise:
@carlapbg Than you for the review.all the inline chnages incorporated.

alebg (approved):

PR #166, opened 2025-04-24
feat(ui-kit): coachingSessionTracker component , storybook and test are added.

### LINKED ISSUE #154: [TSK-959] [TSK-960] ui-kit: CoachingSessionsTracker
[In figma](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=5685-115255&t=2ZDV6HJxfQDBvOSW-4).

Screenshot:

Image

## Notes

- The text 'Coaching sessions included' should be localized
- The text 'Buy more sessions' should be localized; the button callback passed from the outside
- The coaching sessions used here need to extend the [TCoachingOffer model](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/83db9628fed062bbe34c49d1b6b8fce6aaa8d557/packages/models/src/coaching-offer.ts#L24), to include two additional attributes: `used: number` and `included: number`. The latter are the ones that will be shown as "${used}/${included}" (e.g., "0/3", "1/2", and "0/1" in the screenshot), and this component will receive an array of these

manishRise11:
## Description

- Component name: `CoachingSessionsTracker`
- Figma link or screenshot: [Figma Design](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=5685-115255&t=2ZDV6HJxfQDBvOSW-4).

Image

---

## General Checklist

---

## Related Issues

Fixes #154

### DIFF
--- a/packages/translations/src/lib/dictionaries/base.ts
+++ b/packages/translations/src/lib/dictionaries/base.ts
@@ -252,6 +252,12 @@ export const DictionarySchema = z.object({
       rateCallQualityText: z.string(),
       loadMoreText: z.string(),
     }),
+    coachingSessionTracker: z.object({
+      minuteText: z.string(),
+      usedText: z.string(),
+      buyMoreSessionsText: z.string(),
+      coachingSessionText: z.string(),
+    }),
   }),
 
 });
--- a/packages/translations/src/lib/dictionaries/base.ts
+++ b/packages/translations/src/lib/dictionaries/base.ts
@@ -256,7 +256,9 @@ export const DictionarySchema = z.object({
       minuteText: z.string(),
       usedText: z.string(),
       buyMoreSessionsText: z.string(),
+      buyCoachingSessionsText: z.string(),
       coachingSessionText: z.string(),
+      noCoachingSessionText: z.string(),
     }),
   }),
 
--- a/packages/ui-kit/lib/components/coaching-session-tracker/coaching-session-card.tsx
+++ b/packages/ui-kit/lib/components/coaching-session-tracker/coaching-session-card.tsx
@@ -1,0 +1,61 @@
+import { getDictionary, isLocalAware } from "@maany_shr/e-class-translations";
+import { TCoachingOffer } from "packages/models/src/coaching-offer";
+import { FC } from "react";
+import { Badge } from "../badge";
+
+export interface CoachingSessionCardProps extends TCoachingOffer , isLocalAware {
+    used: number;
+    included: number;
+};
+
+/**
+ * A coaching session card component displaying session details and usage tracking.
+ *
+ * @param title - The title/name of the coaching session.
+ * @param duration - Duration of the session in minutes.
+ * @param used - Number of sessions used/consumed.
+ * @param included - Total sessions available in the package.
+ * @param locale - Locale string for internationalization (inherited from `isLocalAware`).
+ *
+ * @example
+ * <CoachingSessionCard
+ *   title="Career Coaching"
+ *   duration={60}
+ *   used={3}
+ *   included={10}
+ *   locale="en"
+ * />
+ *
+ * @remarks
+ * - Uses `Badge` component to display usage ratio (e.g., "3/10 used")
+ * - Integrates with dictionary system for localized text
+ * - Responsive layout with mobile-first sizing
+ */
+
+export const CoachingSessionCard: FC<CoachingSessionCardProps> = ({
+    title,
+    duration,
+    used,
+    included ,
+    locale
+}) => {
+    const dictionary = getDictionary(locale);
+    return (
+        <div className="flex p-2 gap-2 items-center bg-base-neutral-800 border-1 border-base-neutral-700 rounded-medium min-w-[219px] md:w-auto w-full">
+            <div className="flex flex-col items-start gap-1 w-full">
+                <p className="text-sm text-text-primary font-bold leading-[100%]">
+                    {title}
+                </p>
+                <p className="text-sm text-text-secondary leading-[100%]">
+                    {`${duration} ${dictionary.components.coachingSessionTracker.minuteText}`}
+                </p>
+            </div>
+            <Badge 
+                text={`${used}/${included} ${dictionary.components.coachingSessionTracker.usedText}`} 
+                variant="info" 
+                size="big"
+                className="py-1" 
+            />
+        </div>
+    )
+};
--- a/packages/ui-kit/lib/components/coaching-session-tracker/coaching-session-tracker.tsx
+++ b/packages/ui-kit/lib/components/coaching-session-tracker/coaching-session-tracker.tsx
@@ -4,19 +4,19 @@ import { Button } from "../button";
 
 export interface CoachingSessionTrackerProps extends isLocalAware {
     children: React.ReactNode;
-    onClickByMoreSessions: () => void;
+    onClickBuyMoreSessions: () => void;
 };
 
 /**
  * A container component for managing and displaying coaching sessions with a "Buy More" action.
  *
  * @param children - Coaching session cards (typically `CoachingSessionCard` components)
- * @param onClickByMoreSessions - Callback for the "Buy More Sessions" button click
+ * @param onClickBuyMoreSessions - Callback for the "Buy More Sessions" button click
  * @param locale - Locale string for internationalization (inherited from `isLocalAware`)
  *
  * @example
  * <CoachingSessionTracker 
- *   onClickByMoreSessions={handlePurchase} 
+ *   onClickBuyMoreSessions={handlePurchase} 
  *   locale="en"
  * >
  *   <CoachingSessionCard {...session1} />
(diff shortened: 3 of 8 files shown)


manishRise11:
Hello @alebg sir, in this CoachingSessionTracker component, I'm passing the session cards as children. Should I use .map() instead, or is this correct?

alebg:
> Hello @alebg sir, in this CoachingSessionTracker component, I'm passing the session cards as children. Should I use .map() instead, or is this correct?

Hi @manishRise11 . Thanks a lot, that approach is excellent for this case! No need to map props. We'll compose it in the parent, just as you have shown in the story. Looks very good to me overall, but let me give it a further look, and  let's also wait for @carlapbg 's review

packages/ui-kit/lib/components/coaching-session-tracker/coaching-session-card.tsx
alebg, line 1: @manishRise11 please rename this to `coaching-session-item.tsx`, as we're using the name "card" to refer to something else in the project, and it will get confusing

packages/ui-kit/lib/components/coaching-session-tracker/coaching-session-card.tsx
```diff
@@ -0,0 +1,61 @@
+import { getDictionary, isLocalAware } from "@maany_shr/e-class-translations";
+import { TCoachingOffer } from "packages/models/src/coaching-offer";
+import { FC } from "react";
+import { Badge } from "../badge";
+
+export interface CoachingSessionCardProps extends TCoachingOffer , isLocalAware {
```
alebg, line 6: @manishRise11 same, `CoachingSessionItemProps`

packages/ui-kit/lib/components/coaching-session-tracker/coaching-session-tracker.tsx
```diff
@@ -0,0 +1,58 @@
...
+}) => {
+    const dictionary = getDictionary(locale);
+    return (
+        <div className="flex md:flex-row flex-col p-4 gap-4 bg-card-fill border-1 border-card-stroke rounded-medium items-center w-full">
+            <div className="flex flex-col gap-2 items-start w-full">
+                <p className="text-xs text-text-secondary font-bold">
+                    {dictionary.components.coachingSessionTracker.coachingSessionText}
+                </p>
+                <div className="flex gap-2 flex-wrap">
+                    {children}
```
alebg, line 46: @manishRise11 nothing to change here, just to say that the children as props approach here works very nicely. Thanks for thinking this one through!!

packages/ui-kit/lib/components/coaching-session-tracker/coaching-session-tracker.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * @remarks
+ * - Responsive layout (column on mobile, row on desktop)
+ * - Includes localized text for section header and button
+ * - Uses `Button` component with secondary styling for purchases
+ * - Flexible wrapping layout for session cards
+ */
+
+export const CoachingSessionTracker: FC<CoachingSessionTrackerProps> = ({
+    children,
+    onClickByMoreSessions,
```
alebg, line 35: @manishRise11 This is a typo, should read "Buy" not "By". Please rename all variables strings where this typo appears (this is not the only place)

packages/ui-kit/stories/coaching-session-tracker.stories.tsx
```diff
@@ -0,0 +1,97 @@
...
+  ),
+  args: {
+    locale: 'en',
+  },
+};
+
+export const GermanLocale: Story = {
+  render: (args) => (
+    <CoachingSessionTracker {...args} onClickByMoreSessions={() => alert('Buy more coaching sessions')}>
+      {coachingSessions.map((session, index) => (
```
alebg, line 82: @manishRise11 yes very good, this is how we'll use this in the pages!

alebg (changes requested):
@manishRise11 overall it looks very good, thanks a lot! Just please address the minor inline comments I made. Also, could you please add a couple more stories, with more `CoachingSessionItems` to see how do they render when we have a lot of them?

In the figma I don't see what happens if they start to pack. @carlapbg do you think implementing something like a horizontal scroller would be a good idea if the items start getting too close to the edge of the container?

carlapbg:
> @manishRise11 overall it looks very good, thanks a lot! Just please address the minor inline comments I made. Also, could you please add a couple more stories, with more `CoachingSessionItems` to see how do they render when we have a lot of them?
> 
> In the figma I don't see what happens if they start to pack. @carlapbg do you think implementing something like a horizontal scroller would be a good idea if the items start getting too close to the edge of the container?

If there are going to be more items, then I think it would be a good idea to show a second line (that way it’s easier to see and find everything included in the coaching sessions). I don’t think there will be too many items, but just in case that second line fills up, we can add a vertical scroller.

alebg:
> > @manishRise11 overall it looks very good, thanks a lot! Just please address the minor inline comments I made. Also, could you please add a couple more stories, with more `CoachingSessionItems` to see how do they render when we have a lot of them?
> > In the figma I don't see what happens if they start to pack. @carlapbg do you think implementing something like a horizontal scroller would be a good idea if the items start getting too close to the edge of the container?
> 
> If there are going to be more items, then I think it would be a good idea to show a second line (that way it’s easier to see and find everything included in the coaching sessions). I don’t think there will be too many items, but just in case that second line fills up, we can add a vertical scroller.

Right, so more or less like for the skill badges in the coach cards. Perfect then. @manishRise11 could you please implement this? (keeping the children as props strategy, which is very good)

manishRise11:
@carlapbg, ma'am, I have one query regarding the coaching session item. I'm currently using a fixed width of 219px. It looks fine in both 'en' and 'de' as per the attached image. Could you please check and let me know if it’s okay or if I should make any changes?
Screenshot 2025-04-25 at 7 17 47 PM
Screenshot 2025-04-25 at 7 18 12 PM

manishRise11:
Hi @alebg  sir,
I’ve made the following updates in this commit:


- Renamed CoachingSessionCard to CoachingSessionItem.
- Added story for multiple items.
- Resolved the typo error.
- Enabled horizontal scroll for items.

packages/translations/src/lib/dictionaries/base.ts
```diff
@@ -211,6 +211,36 @@ export const DictionarySchema = z.object({
       saveText: z.string(),
       ourPackagesText: z.string(),
     }),
+    sideMenu: z.object({
+      studentText: z.string(),
+      coachText: z.string(),
+      courseCreatorText: z.string(),
+    }),
+    coachingSessionCard: z.object({
```
alebg, line 219: hey @manishRise11 , what are all of these?

alebg (changes requested):
@manishRise11 thanks a lot. Also for adding the empty state story. We didn't think about this case before, and I'm not sure if we will use it at all, but just in case, could you please implement the empty state a bit differently?

So instead of saying 
"Coaching sessions included"
it should say "No coaching sessions included"

and instead of "Buy more sessions", it should say "But coaching sessions"

both translated to German. Apart from that, it looks very good code-wise, but let's wait for @carlapbg 's re-check. In some cases I think the scroll bar is too close to the items though:

immagine

carlapbg:
> @carlapbg, ma'am, I have one query regarding the coaching session item. I'm currently using a fixed width of 219px. It looks fine in both 'en' and 'de' as per the attached image. Could you please check and let me know if it’s okay or if I should make any changes? Screenshot 2025-04-25 at 7 17 47 PM Screenshot 2025-04-25 at 7 18 12 PM

Maybe let's not used a fixed width. I think for the german side it looks kinda weird because all the words are bigger. Every item can have a width of 230px, so it looks better on the german part.

carlapbg (changes requested):
Hi! Apart for the comment I left, can you remove the scrollbar when it's not used? For example in the default option we still see it really close to the items. It should only appear when we have more than 2 lines

manishRise11:
@alebg sir and @carlapbg ma’am, I’ve made the following updates:

- Changed the empty state text from “Coaching sessions included” to “No coaching sessions included.”
- Updated the button label in the empty state from “Buy more sessions” to “Buy coaching session.s”
- Added German translations for the new empty state text
- Increased item width from 219px to 230px to better support German text
- Made the scrollbar visible only when content overflows (dynamic visibility)
- Adjusted scrollbar spacing to prevent overlap with items
- Renamed the callback from ‘onClickBuyMoreSessions’ to ‘onClickBuySessions’

alebg (dismissed):
@manishRise11 Looks good to me, thanks a lot! Let's wait for @carlapbg 's approval and then we're ready to merge

carlapbg (approved):
Looks good to me, excellent work! Thank you

alebg (approved):

Issue #156, opened 2025-04-21
[] [] Small improvements to merged components


alebg:
These are comments from @MytsV , after her work on implementing the HomePage using the ui-kit as it is now. Her work [is in this branch](https://github.com/dream-aim-deliver/dad-e-class-monorepo/tree/demo-rm4). More concretely, around [this file](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/demo-rm4/apps/platform/src/components/home.tsx) .

# Spinner

It is used in the VideoPlayer, but will likely be required in many other parts of UI. I suggest decomposing it into a separate component.

# VideoPlayer

An issue that was already resolved is MuxPlayer having a flicker. It was addressed by making it transparent (or hidden, makes no difference as long as it’s in the tree) until it’s ready.

A similar but less noticeable issue comes up with the thumbnail. With slower internet connection (which we will see in production in comparison to local development) there will be a time gap when the alt text is showing. Using the approach as with MuxPlayer will remove any flickers. The loading could be represented either by a spinner or a skeleton then.

[Video](https://drive.google.com/file/d/1u93GLBUBGIBpNI0uc2U9Dfk2mndQHB6O/view?usp=sharing)

# Tooltip

Flickering happens when the mouse leaves the info icon. This is caused by the way mouse events are handled.     

[[Video](https://drive.google.com/file/d/1F7VbjKFDd6UQ3onEttDVlQszIwtFLoXJ/view?usp=sharing)](https://drive.google.com/file/d/1F7VbjKFDd6UQ3onEttDVlQszIwtFLoXJ/view?usp=sharing)

**21.04.25 Luis Bordo**: You can find Tailwind examples online for tooltips that don't flicker, so we don't need to use React hooks and so on to fix this. Take a look, e.g., at [Flowbite](https://flowbite.com/docs/components/tooltips/).

# Home Accordion

- The accordion container should stretch to the whole width of the screen
- The icon seems not centered with the text and too close to it (not enough padding)

MytsV:
A bit more details for the progress bar issue: to move the slider, clicking really close to it is required, while the thumb is not responsive to clicks. The expected behavior is the slider thumb being clickable and responsive to dragging.

Here's the video showcasing the issue:
[link](https://drive.google.com/file/d/1Gx4G2zQ4H71TPq6GWdrTH9W4uR3guzs4/view?usp=sharing)

PR #168, opened 2025-04-25
Issue improvement to mergerd

ObitoDevv:
**Description :** 

Fix : #156


alebg (changes requested):
@ObitoDevv @anilrise11 thanks a lot! Here's my review, in order of what the issue asked for. I've marked what seems to be fixed, but let's also wait for Viktoriia's review:

- [x] Spinner
- [ ] VideoPlayer: seems to have been ignored
- [ ] Tooltip: you have not explained why my suggestion was not taken into account
- [x] HomeAccordion
- [x]  BuyCoachingSession: cursor-pointer class
- [ ] BuyCoachingSession: onClick with params: could you please add in the stories a callback that just alerts back the `totalCost`? Also, there's a typo in the code
- Carousel/GeneralCard: I'm not so sure if the flickering has in fact been solved, as I also notice it -- I don't know how much flickering is "natural" to have. I'll let @MytsV be the judge of that on this one, so let's wait for her review
- [x] CoachCard: @MytsV for this one, please consider that the minimum width at 348px is already respecting the minimum width as designed in figma
- [x] CoachReviewCard
- [ ] CoachingOnDemandBanner: I still see the flicker as in @MytsV 's video
- [x] Dropdown
- [ ] Hero: It seems that this wasn't addressed at all
- [x] InputField
- [x] NavBar: I'll let Viktoriia judge this one
- [x] ProgressBar

packages/ui-kit/lib/components/carousel/carousel.tsx
```diff
@@ -20,8 +20,8 @@ export const CarouselContent: React.FC<{
       {items.map((item, index) => (
```
MytsV, line 20: Image having a set height works well currently, but there's still an issue with Carousel. @alebg the flicker that you're seeing in the stories is a delay needed to set initial number of items per view.
While listening to resize handler works well after the initial load of the page, there will be a noticeable flicker in integration. We need to find a way to either immediately set the number of items or instead show some placeholder while it's not determined. Might it be a use case for `useLayoutEffect`?

MytsV: Please see [this video](https://drive.google.com/file/d/1A3Ofoz9qParfn7M1DGgNcSkj-lXo5VH2/view?usp=sharing) with reproduction of the issue in the integrated page

MytsV (changes requested):
@ObitoDevv @jatinrise @manishRise11 thank you for the PR! Everything that I didn't comment on looks fine to me.

With the Hero component, here's a hint on why it is causing an overflow: it has both minimum height (300px) an aspect ratio set. The video aspect ratio is 16:9, making the minimum width around 530px. It will cause overflow on many mobile devices.

[video showcase](https://drive.google.com/file/d/17N2rlarEIiUzo7xngSJqbR9U-bORNcds/view?usp=sharing)

packages/ui-kit/lib/components/tooltip.tsx
```diff
@@ -7,115 +7,164 @@ interface TooltipProps {
...
-  const [position, setPosition] = useState<'up' | 'down'>(tipPosition === 'auto' ? 'up' : tipPosition);
-
+function Tooltip({ 
+  text, 
+  title, 
+  description, 
+  contentClassName, 
+  tipPosition = 'auto' 
+}: TooltipProps) {
+  const [isFocused, setIsFocused] = useState(false);
```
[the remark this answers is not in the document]
jatinrise: Hey @alebg @MytsV, thank you for the review!

I noticed that in the `toolkit`, there can be cases where the content overflows outside the viewport — for example:

Screenshot 2025-04-27 at 21 31 24

To address this, I’ve updated the positioning to auto, so now the content will automatically adjust based on the available viewport space:

Screenshot 2025-04-27 at 21 34 00

Going forward, to handle this more robustly, we should add a hook that dynamically measures the position of the content box relative to the viewport and adjusts its placement accordingly.

packages/ui-kit/lib/components/carousel/carousel.tsx
ObitoDevv: Hi @MytsV, thanks for the suggestion! We've updated CarouselController to use useLayoutEffect for setting itemsPerView synchronously, which should eliminate the flicker. The resize listener remains for dynamic updates. @alebg sir, please test in Storybook and integration to confirm. Let me know if further tweaks are needed!

packages/ui-kit/lib/components/navbar.tsx
```diff
@@ -200,7 +200,9 @@ export const Navbar: React.FC<NavbarProps> = ({
 
       {/* Mobile Menu */}
       {isMenuOpen && (
-        <div className="absolute top-0 left-0 w-full h-full bg-button-primary-text text-white flex flex-col items-center justify-center lg:hidden">
+        <div
+          className="fixed top-0 left-0 w-full h-screen bg-button-primary-text text-white flex flex-col items-center justify-center lg:hidden z-9999"
+        >
```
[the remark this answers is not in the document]
ObitoDevv: Oh right ! okay,  I have changed `h-screen` to `h-full`. Now I guess it will work well for you. Thanks for the suggestion . 👍

MytsV:
@ObitoDevv @jatinrise I still unfortunately notice flickering in the carousel in integration. I will be taking over this as well as hero and thumbnail loading, as these are more pronounced and easier to test in the integrated page.

MytsV:
Notes on the fixes:

## Hero

This one is complex, as the current implementation of VideoPlayer and its thumbnail makes it hard to match  Figma. Here the image's dimensions are bounded by height, while in Figma it stretches to the max width.

image
image
image

I made sure that overflow doesn't happen anymore, but @alebg please let me know how critical this issue is and whether we can leave it as is.

## Carousel

The useLayoutEffect here unfortunately didn't give the desired result, as the initial state of 3 items always loaded. This is fixed by relying on the window width when first defining state. But this means this component can't be server-side. I played with integration and this approach along with defining a loading placeholder seems to work fine. We would need to keep this in mind when developing the pages.

On the side note, VideoPlayer component should most definitely also be client-side

```js
import dynamic from 'next/dynamic'

const DynamicComponentWithNoSSR = dynamic(
  () => import('../components/hello3'),
  { ssr: false }
)
```

MytsV:
I propose leaving the other components with the current changes for now. Here are my notes:

## Coaching on demand banner

I feel like instead of using image loading states this way we could rely on Next.js Image component. This can be tried in integration. Also, I am not completely sure if resize handler is a good idea here. Maybe keeping all of the image variants in the DOM with hidden properties based on their breakpoints is better and more performant? Either way, this is best tested in integration, so we can wait until starting developing the pages.

## VideoPlayer

I am also not sure here if passing a Next.js image won't to the thing. If not, an approach like this should be taken
```ts
<div className="absolute inset-0 w-full h-full">
            {!imageLoaded && (
                <div className="absolute inset-0 w-full h-full bg-gray-200 animate-pulse" />
            )}
          <img
            src={thumbnailUrl}
            alt="Thumbnail"
            className="w-full h-full object-contain cursor-pointer"
            onClick={handleThumbnailClick}
            onError={handleThumbnailError}
            onLoad={() => setImageLoaded(true)}
          />
        </div>
```

Most importantly, as a follow-up a skeleton component definitely needs to be developed.

MytsV (dismissed):

alebg (approved):
Thanks a lot guys @MytsV @ObitoDevv @jatinrise @manishRise11 

@anilrise11 Please keep in mind @MytsV 's comments

PR #169, opened 2025-04-25
feat(ui-kit): CourseProgressBar component , storybook and test are added.

manishRise11:
## Description

- Component name: `CourseProgressBar`
- Figma link or screenshot: [Figma Design](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=7607-133014&t=W1BsKllu9WxhzHhp-4).

Screenshot:

Image
---

- [ ] I have avoided **unnecessary re-renders** by using composition (`children` as props), whenever possible. This applies, in particular, to composite components (like lists of components, and the like).
- [ ] **I have designed my components as server-side whenever possible**—I have not used `useState`, `useEffect`, `useContext`, or any other React hook that works with "use client" without a good reason.
- [ ] **I have not put props in state**—in general, props should remain as props unless the component needs an `initialValue`, and judiciously manage its own independent state detached from the parent component.
    
- [ ] **I have avoided using Context** in any critical component in terms of performance

## General Checklist

- [ ] I have run the lint, build, and test commands locally, as explained in the README, before creating this PR to ensure the code is clean and functional.
- [ ] My changes align with the **core models** that have been discussed and agreed upon.
- [ ] The component design has been compared to the **Figma design** to ensure consistency.
- [ ] The **component is fully responsive** and behaves as expected across different screen sizes.
- [ ] The component **supports keyboard web accessibility** (if applicable).
- [ ] I have written **unit tests** or **integration tests** to cover the component's functionality above the coverage threshold.
- [ ] I have updated the **documentation** for the component, including usage examples and guidelines.
- [ ] I have **commented** my code, particularly in complex or non-obvious areas.
- [ ] The **component is properly exported** for use in other parts of the application.

---

## Related Issues

Fixes # (issue)


packages/ui-kit/stories/course-progress-bar.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+  parameters: {
+    layout: 'centered',
+  },
+  argTypes: {
+    locale: {
+      control: 'select',
+      options: ['en', 'de'],
+    },
+    percentage: {
+      control: { type: 'range', min: 0, max: 100, step: 1 },
```
alebg, line 17: @manishRise11 very cool detail for the story!

alebg (dismissed):
@manishRise11 thanks a lot, this is very good! And very nice detail in the storybook with the slider, I didn't know that was possible. Let's wait for @carlapbg 's review and then we are ready to merge

carlapbg (changes requested):
Hi! Just one thing: When using the german version, the progress bar looks way smaller than the english one. I know maybe is because the word in german is way too big. Can we have a fixed size of the progress bar, and depends on how long is the word used in the button the container can grow to the right. I don’t think they are going to use more than 1 word for that button anyway.

Screenshot 2025-04-26 at 20 17 48

Screenshot 2025-04-26 at 20 18 08

manishRise11:
> Hi! Just one thing: When using the german version, the progress bar looks way smaller than the english one. I know maybe is because the word in german is way too big. Can we have a fixed size of the progress bar, and depends on how long is the word used in the button the container can grow to the right. I don’t think they are going to use more than 1 word for that button anyway.
> 
> Screenshot 2025-04-26 at 20 17 48 Screenshot 2025-04-26 at 20 18 08

@carlapbg, Mam, thanks for the feedback! I've fixed the ProgressBar width to 128px to ensure consistency across languages, and the container will now grow based on the button text.

carlapbg:
Thanks! I just notice that the progress bar will get smaller at 10% and even more smaller at 100%. Is there a way it can be always the same size no matter how many numbers the % have?


https://github.com/user-attachments/assets/c4e884af-4334-49f7-b410-4d870a8cfe65

manishRise11:
> Thanks! I just noticed that the progress bar gets smaller at 10% and even smaller at 100%. Is there a way to keep it the same size no matter how many numbers the percentage has?
> 
>  Screen.Recording.2025-04-28.at.01.45.01.mov

@carlapbg  mam, Thanks for pointing that out! The issue happened because the width was adjusting based on the length of the percentage text (like 10% vs 100%). I've fixed it now, so the progress bar stays the same size. Could you please check and confirm if it's working fine for you?

carlapbg (approved):
Thanks @manishRise11 ! Looks good now!

Issue #152, opened 2025-04-21
[TSK-951] [TSK-952] [TSK-998] ui-kit: PreCourseAssessment


alebg:
[In Figma](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=7671-249384&t=W1BsKllu9WxhzHhp-4).

Has two variants:

- Form: 
![Image](https://github.com/user-attachments/assets/2e4a4956-af02-4b80-8b64-7559f2b7ecfb)

- View: 
![Image](https://github.com/user-attachments/assets/eb12b0ea-f873-4363-b29b-7af12c85cc74)

Appears in:
- Student: Pre Course Assessment

Notes:
- Title will be localized by the ui-kit
- The text "Once you submit your response, you won’t be able to make changes, so please review it carefully. Completing this form is required to begin the course." will be localized in the transations package
- The button text ("Send reply")  will be localized by the translations package
- The form variation will have a load state while the callback is being executed, and an error state. These two will be decided by a simple bool prop each (isLoading and isError) passed from the outside. The callback should also be passed from the outside

Props:

This one is a bit complex, so we need to sync-up to explain it better. Essentially, this form can be built in the CMS similar to how course creators build courses, and in fact they share "(form) components".

This is the pseudo-code for the types we may receive. Note that the form variation props are more like a skeleton, of how the form looks like and what it has inside, in order; whereas the view variation has these form components populated (by the input of the student):

Form:
```ts
Component =
  | {
      type: "richText";
      content: string;
      order: number;
    }
  | {
      type: "heading";
      header: string;
      order: number;
    }
  | {
      type: "textInput";
      helperText: string;
      order: number;
    }
  | {
      type: "singleChoice";
      title: string;
      order: number;
      options: {
        name: string;
      }[];
    }
  | {
      type: "multipleChoice";
      title: string;
      order: number;
      options: {
        name: string;
      }[];
    }
  | {
      type: "oneOutOfThree";
      tableTitle: string;
      order: number;
      rows: {
        rowTitle: string;
        columns: {
          columnTitle: string;
        }[];
      }[];
    };

FormProps {
  isLoading: boolean;
  isError: boolean;
  onSubmit: (components: Component[]) => void;
  components: Component[];
}
```

View
```ts
PopulatedFormComponent =
  | {
      type: "richText";
      content: string;
      order: number;
    }
  | {
      type: "heading";
[...]

PR #170, opened 2025-04-25
[TSK-951] [TSK-952] [TSK-998] ui-kit: PreCourseAssessment

jatinrise:
**Description:**
This is a draft PR to demonstrate my approach to building the PreAssessment component.

Summary of Work:
In the `formbuilder` Storybook, I’ve added an array of objects that defines the structure used to dynamically generate the form.

Inside the UI Kit, under `ui-kit/pre-assessment/index.ts`, I’ve created a template that outlines the structure and rendering logic for each form element.

@alebg, could you please review this PR and share your feedback?


packages/ui-kit/lib/components/pre-assessment/form-builder.tsx
```diff
@@ -0,0 +1,70 @@
+import { getDictionary, isLocalAware } from "@maany_shr/e-class-translations";
+import { preAssessmentElements, preAssessmentInstance } from "./index";
+import { useRef, useState } from "react";
+import { Button } from "../button";
+import Banner from "../banner";
+import { IconLoaderSpinner } from "../icons/icon-loader-spinner";
+
+interface FormBuilderProps extends isLocalAware {
+    isLoading: boolean;
+    isError: boolean;
```
alebg, line 10: @jatinrise this isn't doing anything yet, but you'll use it, right?

packages/ui-kit/lib/components/pre-assessment/index.ts
alebg, line 1: @jatinrise this content is very unconventional for an `index.ts` file, which we use normally to declare exports. Please find a different way to organize your code

packages/ui-kit/lib/components/pre-assessment/form-builder.tsx
```diff
@@ -0,0 +1,70 @@
...
+    locale
+}: FormBuilderProps) {
+    const dictionary=getDictionary(locale);
+
+    const formValues = useRef<{ [key: string]: string }>({});
+
+    const submitValue=(id:string,value:string)=>{
+        formValues.current[id] = value;
+    };
+    const handleSubmit = (e) => {
```
alebg, line 29: @jatinrise this `e` is untyped. What is it supposed to represent?

packages/ui-kit/lib/components/pre-assessment/form-builder.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { IconLoaderSpinner } from "../icons/icon-loader-spinner";
+
+interface FormBuilderProps extends isLocalAware {
+    isLoading: boolean;
+    isError: boolean;
+    onSubmit: (PreAssessmentElement:Record<string,string>) => void;
+    elements: preAssessmentInstance[];
+   
+}
+export function FormBuilder({
```
alebg, line 15: @jatinrise This is not the form builder, but the form itself. Form builder is what we'll have in the CMS when we implement it

alebg: @jatinrise you can call this `FormRenderer` for example

packages/ui-kit/lib/components/pre-assessment/index.ts
```diff
@@ -0,0 +1,42 @@
...
+import  singleChoiceElement  from "./single-choice";
+import  textInputElement  from "./text-input";
+
+export type elementType = "richText" |"singleChoice" | "textInput";
+
+export type submitFunction = (key: string, value: string) => void;
+
+export type preAssessmentElement={
+    type:  elementType;
+    designerBtnElement:{
```
alebg, line 12: @jatinrise please add a docstring somewhere explaining what each of these (`designerBtnElement`, `designerComponent`, `formComponent`, `submissionComponent`) are supposed to be and where they are meant to be used

packages/ui-kit/lib/components/pre-assessment/form-builder.tsx
```diff
@@ -0,0 +1,70 @@
+import { getDictionary, isLocalAware } from "@maany_shr/e-class-translations";
+import { preAssessmentElements, preAssessmentInstance } from "./index";
+import { useRef, useState } from "react";
```
alebg, line 3: @jatinrise unused useState forces this to be client-side unnecessarily

packages/ui-kit/lib/components/pre-assessment/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    }>;
+    submissionComponent:React.FC<{
+        elementInstance:preAssessmentInstance
+    }>;
+}
+
+export interface preAssessmentInstance{
+    id:string;
+    type:  elementType;
+    extraAttributes?:Record<string,any>;
```
alebg, line 31: @jatinrise while overall your approach is OK, in particular trying to unify all three uses for this component (CMS form-builder; form renderer; view) this line in particular makes the whole thing badly organized. In general, we shouldn't type things with `any` unless there is a really strong reason, and I see none here. As explained in the issue, we can very well type each type of element using a discriminated union on `type`, such that everything is correctly typed without `any` or `optional` values.

We've been being a bit lax while checking for this in the past components because they are simpler, but in these more complex components we need to keep discipline, otherwise it's going to be very challenging to use and maintain this code. In my opinion, you don't need to black-box an optional `extraAttributes` if you use the discriminated union, which I've been trying to explain in multiple calls, and also on friday.

In the issue itself, you find a very concrete example of how this approach looks like. Taking from the issue, you simply define something like `Element` (or `Component`, as called in the issue):

pseudo-code:
```ts
Element =
  | {
      type: "richText";
      content: string;
      order: number;
    }
  | {
      type: "heading";
      header: string;
      order: number;
    }
  | {
      type: "textInput";
      helperText: string;
      order: number;
    }
  | {
      type: "singleChoice";
      title: string;
      order: number;
      options: {
        name: string;
      }[];
    }
  | {
      type: "multipleChoice";
      title: string;
      order: number;
      options: {
        name: string;
      }[];
    }
  | {
      type: "oneOutOfThree";
      tableTitle: string;
      order: number;
      rows: {
        rowTitle: string;
        columns: {
          columnTitle: string;
        }[];
      }[];
    };
```

As a big logical `OR`, in which all type structures inside the `OR` share a common attribute that defines them (in this case, `type`, which is a Literal). This is what it means to have "a discriminated union on the attribute `type`". You could even structure it as a zod schema like we've been doing in the models (see [here](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/567bd8c6674d627754d1f161bceb86f3ed587764/packages/models/src/profile.ts#L34), for example).

So once you have this, *you don't need optional props*, and you can directly trust that if the `element.type` is `x`, then it will only have the attributes is supposed to have:

```ts
FormRendererProps = {
  ...
  elements: Element[];
}
```

I will let you figure out where exactly to implement the discriminate union approach, since your code is a bit complicated, but again, that approach will prevent you from using `any` or optionals.

Further reason to avoid optionals:
- When using your code, I will need to figure out in which cases I need which optional attribute depending on the case and assert that it actually exists, which makes it a bad developer experience

Further reason to avoid `any`:
- We have no guarantees on what is supposed to happen, which defeats the whole purpose of using typing / TypeScript completely

alebg (changes requested):
@jatinrise @anilrise11 please take a look in detail to my description in the issue and the structure of props I made. Given the complexity of this component, I see very strong arguments *against* the use of `any` in the typing anywhere here, and the use of optional props. See the big inline comment I made for more details. If something is not clear, we can discuss it on Monday

packages/ui-kit/lib/components/pre-assessment/form-builder.tsx
jatingodnani: Yess.

jatinrise:
Hey @alebg, thank you for the review. I have incorporated all the changes.

1. First of all, I have removed the `extraAttributes` and added custom types for each prop as per the defined props above. However, I have one use case for `extraAttributes`: if we need to add properties like `required: true`, we can store them in extraAttributes.
2. I have also changed the file name and updated the overall structure of the file.

Please have a look and let me know your feedback.

packages/ui-kit/lib/components/single-choice-preview.tsx
alebg, line 1: @jatinrise What are this file and component supposed to do?

packages/ui-kit/lib/components/pre-assessment/form-builder.tsx
alebg, line 1: @jatinrise Any reason why this fail, and the stuff inside are still called "builder"? I already pointed this out but you ignored my comment

packages/ui-kit/lib/components/pre-assessment/form-builder.tsx
```diff
@@ -0,0 +1,74 @@
...
+            <Banner
+                style="warning"
+                description={dictionary.components.formBuilder.alertText}
+            />
+            <div className="relative w-full">
+                <Button
+                    type="submit"
+                    variant="primary"
+                    className="w-full capitalize"
+                    disabled={isLoading || isError}
```
alebg, line 67: @jatinrise please remove the `isError` condition here. Reasoning: if there's an error, we should be able to actually try again. Otherwise we're prevented from clicking the button ever again unless we refresh, which makes us lose the data we inputted in the form

packages/ui-kit/lib/components/pre-assessment/text-input.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  return (
+    <div className="text-text-primary flex flex-col gap-2">
+      <RichTextEditor
+        locale="en"
+        name={`rich-text-${elementInstance.id}`}
+        placeholder="Start typing..."
+        initialValue={value}
+        onChange={(value) => setValue(value)}
+        onLoseFocus={onLoseFocus}
```
alebg, line 49: @jatinrise Just to check, will `onLoseFocus` always trigger when we click on the submit button of the form?

packages/ui-kit/lib/components/pre-assessment/text-input.tsx
```diff
@@ -0,0 +1,67 @@
...
+    if (!submitValue || !value) return;
+    const content = serialize(value);
+    submitValue(elementInstance.id.toString(), content);
+  };
+
+  return (
+    <div className="text-text-primary flex flex-col gap-2">
+      <RichTextEditor
+        locale="en"
+        name={`rich-text-${elementInstance.id}`}
```
alebg, line 45: this shouldn't have `rich-text`in its name, should it?

packages/ui-kit/lib/components/pre-assessment/text-input.tsx
```diff
@@ -0,0 +1,67 @@
...
+        placeholder="Start typing..."
```
alebg, line 46: @jatinrise Please don't hard code English values in the UI

packages/ui-kit/lib/components/pre-assessment/text-input.tsx
```diff
@@ -0,0 +1,67 @@
...
+    </div>
+  );
+}
+
+function ViewComponent({ elementInstance }: { elementInstance: FormElement }) {
+  if (elementInstance.type !== FormElementType.TextInput) return null;
+
+  return (
+    <div className="text-text-primary flex flex-col">
+      <p>{elementInstance.helperText}</p>
```
alebg, line 60: @jatinrise please check [my answer to your question in the issue](https://github.com/dream-aim-deliver/dad-e-class-monorepo/issues/152#issuecomment-2832284470), where I try to clarify the difference between inputText.helperText and inputText.content. Here both things are mixed. Let me know if my answer is not enough to clarify what's supposed to happen

alebg: @jatinrise Also, both content and helperText are rich text, so they need the RichTextRenderer to be viewed properly

packages/ui-kit/stories/assessment-form.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+      return {
+        question: "Sample multiple choice question?",
+        options: ["Option A", "Option B", "Option C"]
+      };
+    default:
+      return { content: "Default content" };
+  }
+}
+
+export default {
```
alebg, line 40: @jatinrise this is broken in storybook:
immagine

packages/ui-kit/lib/components/multiple-choice-preview.tsx
alebg, line 1: @jatinrise also this file and the things inside, what are they supposed to be?

packages/ui-kit/lib/components/banner.tsx
alebg, line 1: @jatinrise please put this file in the pre-assessment folder, as we have many things named "banner" already and it will be confusing

packages/ui-kit/lib/components/pre-assessment/types.ts
```diff
@@ -0,0 +1,311 @@
...
+ * @param value - The submitted value from the form element
+ * 
+ * @example
+ * ```tsx
+ * const handleSubmit: SubmitFunction = (key, value) => {
+ *   console.log(`Element ${key} submitted with value:`, value);
+ * };
+ * ```
+ */
+export type SubmitFunction = (key: string, value: any) => void;
```
alebg, line 40: @jatinrise Could you please explicitly justify why we have an `any` here?

alebg: @jatinrise Again, this could be a family of functions depending on the element type, and then this general form could be some sort of type union on these functions. But let me know if there's a strong argument for not doing this here. Again, I can't think of any

alebg (changes requested):
@jatinrise thanks, it looks much better now. Please address my inline comments

Also, I'm not sure if I was supposed to review the single choice element yet, but in case you haven't noticed, something is not working with the wiring:

If I don't select any option, then the final values passed to the outside from the form don't contain the single choice object at all:

immagine

And if I do choose something, the values tell me that no option was picked (all have `isSelected = false`):
immagine

alebg:
> @jatinrise @anilrise11 please take a look in detail to my description in the issue and the structure of props I made. Given the complexity of this component, I see very strong arguments _against_ the use of `any` in the typing anywhere here, and the use of optional props. See the big inline comment I made for more details. If something is not clear, we can discuss it on Monday

Hi @jatinrise @anilrise11,
I still notice at least one instance of `any` in the typing of this component (check my inline comment above). As I mentioned earlier, while we *can* in general make exceptions to our guidelines after discussing together, for complex components like this one we would need a strong and explicit justification on your part to do so.

Also, before you mark something as ready for re-checking, could you please make sure all previous feedback has been fully addressed? This helps avoid having to revisit points multiple times and ensures we can move more efficiently through the review process. Thanks for understanding and for your efforts on this.

anilrise11:
> > @jatinrise @anilrise11 please take a look in detail to my description in the issue and the structure of props I made. Given the complexity of this component, I see very strong arguments _against_ the use of `any` in the typing anywhere here, and the use of optional props. See the big inline comment I made for more details. If something is not clear, we can discuss it on Monday
> 
> Hi @jatinrise @anilrise11, I still notice at least one instance of `any` in the typing of this component (check my inline comment above). While we _can_ in general make exceptions to our guidelines after discussing together, as I mentioned earlier, for complex components like this one we would need a strong and explicit justification on your part to do so.
> 
> Also, before you mark something as ready for re-checking, could you please make sure all previous feedback has been fully addressed? This helps avoid having to revisit points multiple times and ensures we can move more efficiently through the review process. Thanks for understanding and for your efforts on this.

Hello @alebg Yes I agree with you we should not be using "any" unless we have strong explanation for it, we are incorporating the changes discussed in the above comments, better these changes & approaches have come in the draft PR, it helps early on  in case of complex components.....

jatinrise:
Hey @alebg , below are the changes included in the above commit:

1. Ensured proper validation: If a field is marked as required, a localized error message will now be displayed when the form is submitted.
2. Fixed a bug related to the single choice input component.
3. Added the `pre-assessment/view` story to showcase the view form, and included view mode support for all related components.
4. Created a lesson-component folder to centralize common lesson components used across both `CMS` and `CourseBuilder`.
5. Introduced a designer component layout component to serve as a template structure.

Work in progress:

 

- [ ] Currently working on the designer component for each lesson, along with the changes we discussed yesterday.

Let me know if the validation behavior and error message localization look good to you, or if any further adjustments are needed.

packages/translations/src/lib/dictionaries/base.ts
```diff
@@ -230,6 +230,17 @@ export const DictionarySchema = z.object({
       rateCallQualityText: z.string(),
       loadMoreText: z.string(),
     }),
+    formBuilder: z.object({
```
alebg, line 233: @jatinrise another detail, this is not the "formBuilder" but the form renderer

packages/ui-kit/lib/components/pre-assessment/types.ts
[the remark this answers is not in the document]
alebg: @jatinrise , as a remark, a part of this file for sure should go to lesson-components, but it *could* be that there's specific things for the pre-assessment form. Those should remain in this folder

alebg (changes requested):
@anilrise11 @jatinrise thanks a lot, this is looking pretty good! My inline comments are minor, but please address them. Also, while this looks good in the abstract, it would be nice if you could implement a couple more complex stories so I can have a more concrete implementation to analyze.

Here's a couple of issues I noticed in the story that works for me, though:

immagine

jatinrise:
> @anilrise11 @jatinrise thanks a lot, this is looking pretty good! My inline comments are minor, but please address them. Also, while this looks good in the abstract, it would be nice if you could implement a couple more complex stories so I can have a more concrete implementation to analyze.
> 
> Here's a couple of issues I noticed in the story that works for me, though:
> 
> immagine

Hey @alebg, the output will be shown based on which lesson the student fills out first.

packages/ui-kit/lib/components/lesson-components/designer-layout.tsx
```diff
@@ -0,0 +1,94 @@
+import React from "react";
+import { getDictionary, isLocalAware } from "@maany_shr/e-class-translations"
+import { IconTrashAlt } from "../icons/icon-trash-alt"
+import { IconButton } from "../icon-button"
+import { IconChevronUp } from "../icons/icon-chevron-up"
+import { IconChevronDown } from "../icons/icon-chevron-down"
+/**
+ * Designer Layout Component
+ * This component provides the layout structure for the form designer interface.
```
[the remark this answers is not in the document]
jatinrise: Hey @alebg , it's a general term that applies to both, since there are common elements in the Course Builder and CMS, so I created this layout.

jatinrise:
Hey @alebg, below are the changes included in the above commit:

1. Completed the three lesson components, each with `designer,` `view,` and `edit` subcomponents.
2. Incorporated all inline changes and created a `pre-assessment` folder inside Storybook to showcase all pre-assessment components.
3. Added an animated button (designer component) that allows the admin to toggle whether a field is required.

packages/ui-kit/lib/components/lesson-components/single-choice.tsx
```diff
@@ -0,0 +1,185 @@
...
+
+    const [options, setOptions] = useState<optionsType[]>([]);
+
+    useEffect(() => {
+        const mappedOptions = elementInstance.options.map(option => ({
+            ...option,
+            isSelected: false
+        }));
+        setOptions(mappedOptions);
+    }, [elementInstance.options]);
```
MytsV, line 128: We definitely need to initialize the options in the useState, or else we get a flicker
If elementInstance contains the options and gets updated on submitValue, why shouldn't we just use it without a separate state?

And maybe singleChoiceValue type? So that they match and we don't require any mapping. Undefined isSelected means false then.

packages/ui-kit/lib/components/pre-assessment/form-renderer.tsx
```diff
@@ -0,0 +1,193 @@
...
+                            />
+                            {formErrors[elementInstance.id] && (
+                                <Banner
+                                    style="error"
+                                    description={getErrorMessage(elementInstance)}
+                                />
+                            )}
+                        </div>
+                    );
+                })}
```
MytsV, line 169: Just to note - we don't sort elements by their order here. I am fine with doing this outside the form renderer, but we should keep it in mind

MytsV (changes requested):
Hi @jatinrise! Here's my review of the components. There are a few visual bugs and some parts that require clarification.

packages/ui-kit/lib/components/lesson-components/designer-layout.tsx
```diff
@@ -0,0 +1,131 @@
...
+                            value="required"
+                            checked={isChecked}
+                            size="small"
+                            onChange={onChange}
+                          
+                            
+                        
+                        />
+                        </div>
+                    }
```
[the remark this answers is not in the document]
jatinrise: Heyy @MytsV, there are more cases where we don't need to show the animated button, such as in course-builder lessons, rich-text, and heading-lesson.

alebg:
@jatinrise as discussed in the call on Monday, please differentiate the types for the different cases (view, form, form builder), as opposed to using optionals

packages/ui-kit/lib/components/lesson-components/single-choice.tsx
jatinrise: Hey @MytsV, if we don’t manage the state, we won’t be able to see the changes. Also, submitValue doesn’t change the element’s instance state—it only collects the value from the single choice.

jatinrise:
Hey @alebg, I’ve completed all the pre-assessment components. Should I create a new PR to make the review process easier?

alebg:
> Hey @alebg, I’ve completed all the pre-assessment components. Should I create a new PR to make the review process easier?

Yes please!

alebg:
Closed in favor of #191

PR #187, opened 2025-05-03
fix(ui-kit):Fix upload component to support displaying multiple files and images.

### LINKED ISSUE #186: fix(ui-kit): Fix upload component to support displaying multiple files and images.
As pointed out by @alebg, we need to make sure to add `onImageUpload` and `onFileUpload` as callbacks in the component.

Image

jatinrise:
- Component name: `image-uploader` `file-uploader` `video-uploader`

This PR includes the following changes:

1. Separated upload logic into three different files: image, file, and video, as each requires different props. Additionally, multiple file uploads are now supported with a configurable upload limit.
2. Made some updates to the tests.

Screenshot 2025-05-03 at 14 33 46

## General Checklist

fix #186


alebg (changes requested):
@jatinrise @anilrise11 So I see that most of the basic code comes from an old PR we merged. So I think it is a good time to improve it. Please take my comments below as a revision of what was already done, but being more strict regarding our guidelines, in particular given that this will be used in a crucial part of the app.

So, in general, the typing of what was already merged is rather opaque and confusing, and adding new things on top of this starts to feel inconsistent. At this point, I'd say that what was done needs a re-implementation (and all the places where used a refactor, as needed), as opposed to piling new code on top. So here's more concrete comments:

- Why is the name of the upload props "Enhanced"? If not meaningful, please omit naming things in a confusing way

- I don't agree with the `type` prop being optional. We've discussed that making props unnecessarily optional makes it a bad developer experience when trying to use the ui-kit

- To add up to the previous point, when you have something like a type prop, a better approach is to use a union type strategy.  For example, right now the "single" type of file uploader can receive a "maxFiles" prop, which doesn't make conceptual sense. If you were to use a union type instead, we wouldn't need to pass irrelevant props in different logical cases. Same with the callbacks, I don't see why they should be optional

- I don't remember now why we approved the `text` prop is a strategy. Couldn't we just take those values from the dictionary? If it's because the text shown per category changes per lesson component, for example, then we can pass it from the outside, but then I don't see why would these fields be optional, whereas the `text` itself is not

- The download functionality is technically application logic, not ui-kit logic. It should be injected as a callback from the outside. The idea is that the app will decide what's the best way to handle downloads, as opposed to the UI kit enforcing this, and this will be more crucial in the course part of the app

- The upload callback should be a single one in my opinion, I don't see a reason to have two

packages/ui-kit/lib/components/drag&drop/file-uploader.tsx
```diff
@@ -0,0 +1,218 @@
...
+ * Represents a file being uploaded with its associated state
+ */
+export type FileUploaderType = {
+  file: File;
+  isUploading: boolean;
+  error?: boolean;
+  serverData?: UploadResponse;
+};
+
+export type EnhancedUploadFileProps = {
```
alebg, line 30: confusing naming. Why "Enhanced"?

jatinrise:
This commit includes the following changes:

- Updated ImageUploader and VideoUploader components to use a unified props structure for single and multiple uploads.
- Integrated localization support using getDictionary for text elements in the uploaders.
- Enhanced file upload handling with improved error management and state updates.
- Modified the ProfessionalInfo and ProfileInfo components to align with the new uploader structure.
- Updated storybook stories for `FileUploader`,` ImageUploader`, and `VideoUploader` to include locale options and better demonstrate functionality.
- Added tests for DragAndDrop component to validate file size restrictions and error handling.

jatinrise:
Hey @alebg, I'm not able to see the inline comment in my file, which is why I’ve added the screenshot below. So, I'll provide a point-wise explanation for each comment instead.
Screenshot 2025-05-05 at 11 45 14 AM (1)

1. Yes, you're right that the type props shouldn't be optional. That's why I removed the optional properties and also used a union type for each prop.
2. have removed the` onDownload `function from the UI kit and added it as a callback.
3. Previously, I was passing two props: onUpload and onImageUpload callbacks. Now, I’ve combined them into a single callback called `onImageUpload`, which is no longer optional.
4. I have removed the `text` as a prop and localized the hardcoded text using the dictionary.

jatinrise:
Hey @alebg, the above commit includes the following changes:

1. I’ve created a single component to handle all three cases (file, image, and video) using a variant prop. You can specify the accepted file types using the acceptedFiles prop as an array.
2. To give a better idea of how the loading state will look, I’ve added a setTimeout to simulate a delay and show the UI.
3. I’ve also added error handling and its corresponding UI.

alebg:
Hi @jatinrise, thanks. Just a remark before I review your latest changes, this PR is very messy now. Maybe because you did a merge commit and solved conflicts?

image

With a lot of files being changed:

image

But it's hard to tell just like that if all the changes in this PR to the non-relevant files to this task are keeping what was already done. Could you please create a new clean branch and put your work there?

In the future, as explained since the beginning, the preferred way to update your branches is through `git rebase main`. Please keep this in mind.

Also, unless you need something that was merged into main, you can just not do any merge commits or rebasing and let me handle this once I finish reviewing, that's how we've been operating the past couple of weeks.

PR #195, opened 2025-05-12
[TSK-978] [TSK-979] ui-kit: ComponentImage, ComponentVideo, ComponentImageCarousel #172

### LINKED ISSUE #172: [TSK-978] [TSK-979] ui-kit: ComponentHeading, ComponentImage, ComponentVideo, ComponentImageCarousel

jatinrise:
Component name: `Coursebuilderlesson`

- Added file size tracking to UploadedFileType and related response types.
- Updated FilePreview component to display uploading status with a badge.
- Implemented new Designer and Form components for image files, image galleries, and video files.
- Created stories for image files, image galleries, and video files to showcase new components.
- Refactored course builder types to include new file types and their properties.

Fix #172

### DIFF
--- a/packages/ui-kit/lib/components/course-builder-lesson-component/image-files.tsx
+++ b/packages/ui-kit/lib/components/course-builder-lesson-component/image-files.tsx
@@ -62,7 +62,7 @@ export function DesignerComponent({ elementInstance, locale, onUpClick, onDownCl
     return (
         <DesignerLayout
             type={elementInstance.type}
-            title="Image"
+            title={dictionary.components.courseBuilder.ImageFileText}
             icon={<IconImage classNames="w-6 h-6" />}
             onUpClick={() => onUpClick(elementInstance.id)}
             onDownClick={() => onDownClick(elementInstance.id)}
--- a/packages/ui-kit/lib/components/course-builder-lesson-component/image-files.tsx
+++ b/packages/ui-kit/lib/components/course-builder-lesson-component/image-files.tsx
@@ -1,12 +1,19 @@
-import { useState } from "react";
-import { courseElement, CourseElementTemplate, CourseElementType, DesignerComponentProps, FormComponentProps } from "../course-builder/types";
+
+import { CourseElementTemplate, CourseElementType, DesignerComponentProps, FormComponentProps } from "../course-builder/types";
 import DesignerLayout from "./designer-layout";
 import { getDictionary } from "@maany_shr/e-class-translations";
-import { IconCloudDownload } from "../icons/icon-cloud-download";
 import { UploadedFileType, Uploader, UploadResponse } from "../drag-and-drop/uploader";
-import { ImageFilePreview } from "./types";
 import { IconImage } from "../icons/icon-image";
+/**
+ * @fileoverview Image file component for the course builder.
+ * This file contains components and types for handling image uploads and display
+ * in the course builder interface.
+ */
 
+/**
+ * Configuration object for the image file element in the course builder.
+ * Defines the type, button appearance, and component references for image handling.
+ */
 const imageFilesElement: CourseElementTemplate = {
     type: CourseElementType.ImageFile,
     designerBtnElement: {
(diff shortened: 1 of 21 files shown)


jatinrise:
Hi @alebg, I've marked this PR as a draft since Manish is currently working on the same Course Builder lesson. Once his PR is merged, I'll sync my changes accordingly. In the meantime, I'd appreciate it if you could review the code.

packages/ui-kit/lib/components/course-builder-lesson-component/image-files.tsx
```diff
@@ -0,0 +1,102 @@
...
+            onDeleteClick={() => onDeleteClick(elementInstance.id)}
+            locale={locale}
+            courseBuilder={true}
+        >
+            <Uploader
+                type="single"
+                variant="image"
+                file={file}
+                onFilesChange={handleFilesChange}
+                onDelete={handleDelete}
```
alebg, line 78: @jatinrise I don't understand why onDelete and onDownload are doubled? this looks too confusing as an implementation, but let me know if there's a good reason

packages/ui-kit/lib/components/course-builder-lesson-component/image-files.tsx
```diff
@@ -0,0 +1,102 @@
...
+        setFile(null);
+        return Promise.resolve({
+            imageId: '',
+            imageThumbnailUrl: '',
+            fileSize: 0
+        });
+    };
+
+    const handleDelete = () => {
+        setFile(null);
```
alebg, line 55: @jatinrise I don't think removing the file from the ui-kit is the best. For example, if the delete API call fails, it would seem to the user that it was successful since the file is not there. If this kind of logic is handled by the page instead, then there we can decide better when to visually remove the file

packages/ui-kit/lib/components/course-builder-lesson-component/image-files.tsx
```diff
@@ -0,0 +1,102 @@
...
+        icon: IconImage,
+        label: "Image"
+    },
+    designerComponent: DesignerComponent,
+    formComponent: FormComponent
+};
+
+export function DesignerComponent({ elementInstance, locale, onUpClick, onDownClick, onDeleteClick }: DesignerComponentProps) {
+    if (elementInstance.type !== CourseElementType.ImageFile) return null;
+    const dictionary = getDictionary(locale);
```
alebg, line 22: @jatinrise this `dictionary` is not being used, why is it here?

packages/ui-kit/lib/components/course-builder-lesson-component/image-files.tsx
```diff
@@ -0,0 +1,102 @@
...
+    );
+}
+
+/**
+ * Form Component for Download Files
+ * Renders the rich text content in the form view
+ */
+export function FormComponent({ elementInstance, locale }: FormComponentProps) {
+    if (elementInstance.type !== CourseElementType.ImageFile) return null;
+    const dictionary = getDictionary(locale);
```
alebg, line 92: @jatinrise this is not being used either

alebg (changes requested):
@jatinrise I don't understand very well, you asked me to review this PR, but a quick check on a single file made me notice that there are many small issues here and there. I know you marked this as a draft, but you asked me to review the code, hence I expect it to be clean enough. I only checked one file but it has unused dictionaries being instantiated, for example. Please clean these details all around the PR, I leave it up to you to check the other files

About the implementation itself, I find it confusing, why are the download and delete functionalities duplicated between the designer layout component and the uploader component?

packages/ui-kit/lib/components/course-builder-lesson-component/image-files.tsx
alebg: @jatinrise  @anilrise11 if you're going to go against the guidelines, could you please at least provide arguments for your reasoning? I gave you a clear example of why enforcing this `handleDelete` with no control from the page is problematic.

We've discussed at length several times that the UI kit should be kept "dummy" and most functionalities will be injected as callbacks. Why would we do it differently for this case?

jatinrise:
Hi @alebg, I’ve identified the problem and will address all the issues.

jatinrise:
Hi @alebg,
I’ve made the changes as per the following rules:

1. Added callbacks to pass values to the parent
2. Removed unused code
3. Added documentation for better clarity

Please review.

alebg:
Thanks @jatinrise . So the approach looks cleaner now. But it still needs some work, right? For example the form stories for the image file don't render:

immagine

jatinrise:
close in favour #199

PR #197, opened 2025-05-14
[TSK-885] [TSK-886] ui-kit: UserGrid

### LINKED ISSUE #131: [TSK-885] [TSK-886] ui-kit: UserGrid
For this, please use the work done by @MytsV in #141. This provides a base grid, over which a basic UserGrid has been implemented.

**IMPORTANT: Please note that we're not exactly following the figma design for this one**. Instead, we're adapting AG Grid to provide the same functionalities in the figma, with a look as close as possible. If trying to match something about the look breaks functionality, generates too many visual bugs, or is too much work, please drop it and we'll implement an alternative.

# UserGridFilterModal

There's no figma for this, but please take inspiration from the course filter modal of the CMS:

Image

[Figma link](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=7252-332368&t=Z9yOThOzKCxx7IT2-4).

Notes:
- The properties that you can select / manipulate are 
- This one just grabs any info you selected / manipulated, and passes it to the callback `onApplyFilters`. 

# UserGrid

Screenshot:

Image

On figma: https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=8784-216191&t=VII2zZycq4mOnDlM-4

Notes:

**IMPORTANT:** Viktoriia has provided a pre-made UserGrid, butyou need to make sure that the points below are implemented. Even if some of them are already there.

- For the data, please extend the `TBasePersonalProfile` model ([link](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/83db9628fed062bbe34c49d1b6b8fce6aaa8d557/packages/models/src/profile.ts#L16C13-L16C33)) and extend it to add the missing columns. Please also add the prop `userId: number` to this extended model
- Please enable the AG Grid default sorting and filtering for all columns (even if this is not shown in the figma)
- The search bar at the top is a fuzzy search across the following fields only: `name`, `surname`, `email, `phone`
- For the batch actions, we'll pass callbacks from the outside and that's it. For now we only have one, which is send notifications. This callback will need to be passed an array of `userId` matching exactly those selected. If no user is selected, then you can show a localized error message on clicking the send notifications button, saying "Error: no users selected to send notification". If this happens, don't execute the callback.
- Needs a checkbox selector per row, and the general one. This is also provided by default AG Grid. You'll get the userIds for the send notification callback from the selected users
- When clicking the "Filters" button, it should call the filters modal. These filters will be purely client side, so the callback `onApplyFilter` can be passed from UserGrid to UserGridFilterModal directly. They will essentially replicate what the filter columns already do

## Instructions from Viktoriia

1. **Check consistency of AG Grid with other UI elements.**

AG Grid is styled in the CSS file. I tried to match the Figma design visually, but it should be 100%. To see possible theming variables, use the [[Theme Builder](https://www.ag-grid.com/theme-builder/)](https://www.ag-grid.com/theme-builder/). For example, Cell Editing Border will correspond to --ag-cell-editing-border variable.

2. **Scrollbars should be globally set to dark mode**

3. **Shimmering skeleton**

The table should be covered while it is in the initial loading state. I did it with some mock grey background, but a better approach is to implement a shimmering skeleton which can be seen in certain Figma components. Please see this example and whether you could copy something from it.

https://ui.shadcn.com/docs/components/skeleton

Image

4. **Style loading and no rows state**

Unfortunately, there is no indication in Figma on how these should be handled specifically for the tables. But maybe there are similar approaches documented that could be used here for consistency. Please let me know if you need direction on overriding these.

Image

Image

ObitoDevv:
## Description

- Component name: `User Grid`
- Figma link or screenshot: [Figma Design](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=7252-332368&t=Z9yOThOzKCxx7IT2-4)

---

## General Checklist

---

## Related Issues

Fixes #131  (issue)


alebg:
Hi @ObitoDevv , as discussed in the call, please integrate the icons I send to @anilrise11 in the UserGrid, as per this screenshot (made by @carlapbg ):

immagine

Please note which icon corresponds to which role. Also, you can modify slighltly the user props such that you always have objects `roles: {platformName: string; role: string}[]` which you can use to render the platform name + icon

carlapbg (changes requested):
When selected, the icon should have the same color as the text. Both should change the same

Screenshot 2025-05-15 at 22 42 26


Also, is it possible to add a "Clear all filters" button? Something like this:
Screenshot 2025-05-15 at 22 55 45

ObitoDevv:
@carlapbg Hi mam, I have fixed the tab color issue and added the `clear filter` button to clear all filters at once. 

Can you please verify the changes ?

MytsV:
I fixed all of the small issues I found. The changes I requested personally seem to be addressed in a different PR.

A few notes:
- there might be a need to indicate the number of filters applied like in Figma
- there's some flicker of the button's color that I see when reloading the page. can't pinpoint the reason for it yet, will have to see in the integration

MytsV (dismissed):

alebg (dismissed):
Thanks a lot @ObitoDevv @MytsV !

@ObitoDevv @anilrise11 See my last commit: note that I refactored a couple of components that were missing localization. Please also take note of the changes made by Viktoriia and me on the UserGrid

packages/ui-kit/lib/components/grids/user-grid.tsx
```diff
@@ -70,69 +186,487 @@ export const UserGrid = (props: UserGridProps) => {
...
-            suppressPaginationPanel={true}
-            paginationAutoPageSize={true}
-            isExternalFilterPresent={() => props.doesExternalFilterPass !== undefined}
-            doesExternalFilterPass={props.doesExternalFilterPass}
-            rowSelection={props.enableSelection ? { mode: 'multiRow' } : undefined}
-        />
-    </div>;
+
+        if (props.onSendNotifications) {
+            alert(dictionaryFormat(dictionary.sendingNotifications, { count: selectedUserIds.length }));
```
alebg, line 305: @maany I added a small util in the translations package to allow me to mimick Python's format strings and do this. This entry in the dictionary is `      sendingNotifications: 'Sending notifications to %{count} selected users',`  and  this util function replaces `%{count}` with the dynamic value.

Downside of the util is that it's not typed or anything, so nothing fails if you don't pass the format keys needed, or if you pass a key that is not present. TBD if this is a good util to have, and if so, how to improve it.

maany: You can do this in js with 
```js
export default const Component = (props: {count: number}) => {
  return (
    <div>
        `Sending notifications to ${props.count}`
    </div> 
}
```

maany: I see, we would need something like liquid js in the translations package in that case. https://liquidjs.com/

PR #198, opened 2025-05-14
[TSK-910] [TSK-911] [TSK-912] NotificationGrid

### LINKED ISSUE #134: [TSK-910] [TSK-911] [TSK-912] NotificationGrid
For this, please use the work done by @MytsV in #141. This provides a base grid, over which a basic NotificationGrid has been implemented.

**IMPORTANT: Please note that we're not exactly following the figma design for this one**. Instead, we're adapting AG Grid to provide the same functionalities in the figma, with a look as close as possible. If trying to match something about the look breaks functionality, generates too many visual bugs, or is too much work, please drop it and we'll implement an alternative.

**In particular, this will look more like a normal AG Grid table than what we see in the figma**.

This appears in the "Activity History" sections of the figma.

# [Figma link](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=8446-188518&t=Z9yOThOzKCxx7IT2-4)

# NotificationGrid

Screenshots:

- Empty state

Image

- Student view

Image

- Coach view (looks the same to me as the student view though, but please check if Stefano has made any difference)

Image

- CMS: note the tabs on top that filter by platform. The names of the platforms have to be passed from the outside as a prop, and then on click of any of the tabs, a filter of that string on the platform attribute of the notification has to be applied

Image

Notes:

**IMPORTANT:** Viktoriia has provided a pre-made NotificationGrid, but you need to make sure that the points below are implemented. Even if some of them are already there.

- For the data, please extend the `TNotification` model ([link](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/83db9628fed062bbe34c49d1b6b8fce6aaa8d557/packages/models/src/notification.ts#L44)) and extend it to add the missing columns. Please also add the prop `notificationId: number` to this extended model, and the `platform: string` prop (which won't be rendered in the non-CMS views)
- Please enable the AG Grid default sorting and filtering for all columns (even if this is not shown in the figma)
- The search bar at the top is a fuzzy search across the following fields only: `message`, `action.title`
- For the "Mark All" batch action, we'll pass a callback from the outside and that's it. This callback will need to be passed an array of `notificationId` matching exactly those selected. If no notification is selected, then you can show a localized error message on clicking the send notifications button, saying "Error: no notifications selected to mark as read". If this happens, don't execute the callback.
- Needs a checkbox selector per row, and the general one. This is also provided by default AG Grid. You'll get the notificationIds for the mark all as read callback from the selected notifications

## Instructions from Viktoriia
For this, please use the work done by @MytsV in #141. This provides a base grid, over which a basic UserGrid has been implemented.

**IMPORTANT: Please note that we're not exactly following the figma design for this one**. Instead, we're adapting AG Grid to provide the same functionalities in the figma, with a look as close as possible. If trying to match something about the look breaks functionality, generates too many visual bugs, or is too much work, please drop it and we'll implement an alternative.

# UserGridFilterModal

There's no figma for this, but please take inspiration from the course filter modal of the CMS:

Image

[Figma link](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=7252-332368&t=Z9yOThOzKCxx7IT2-4).

Notes:
- The properties that you can select / manipulate are 
- This one just grabs any info you selected / manipulated, and passes it to the callback `onApplyFilters`. 

# UserGrid

Screenshot:

Image

On figma: https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=8784-216191&t=VII2zZycq4mOnDlM-4

Notes:

**IMPORTANT:** Viktoriia has provided a pre-made UserGrid, butyou need to make sure that the points below are implemented. Even if some of them are already there.

- For the data, please extend the `TBasePersonalProfile` model ([link](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/83db9628fed062bbe34c49d1b6b8fce6aaa8d557/packages/models/src/profile.ts#L16C13-L16C33)) and extend it to add the missing columns. Please also add the prop `userId: number` to this extended model
- Please enable the AG Grid default sorting and filtering for all columns (even if this is not shown in the figma)
- The search bar at the top is a fuzzy search across the following fields only: `name`, `surname`, `email, `phone`
- For the batch actions, we'll pass callbacks from the outside and that's it. For now we only have one, which is send notifications. This callback will need to be passed an array of `userId` matching exactly those selected. If no user is selected, then you can show a localized error message on clicking the send notifications button, saying "Error: no users selected to send notification". If this happens, don't execute the callback.
- Needs a checkbox selector per row, and the general one. This is also provided by default AG Grid. You'll get the userIds for the send notification callback from the selected users
- When clicking the "Filters" button, it should call the filters modal. These filters will be purely client side, so the callback `onApplyFilter` can be passed from UserGrid to UserGridFilterModal directly. They will essentially replicate what the filter columns already do

## Instructions from Viktoriia

1. **Check consistency of AG Grid with other UI elements.**

AG Grid is styled in the CSS file. I tried to match the Figma design visually, but it should be 100%. To see possible theming variables, use the [[Theme Builder](https://www.ag-grid.com/theme-builder/)](https://www.ag-grid.com/theme-builder/). For example, Cell Editing Border will correspond to --ag-cell-editing-border variable.

2. **Scrollbars should be globally set to dark mode**

3. **Shimmering skeleton**

The table should be covered while it is in the initial loading state. I did it with some mock grey background, but a better approach is to implement a shimmering skeleton which can be seen in certain Figma components. Please see this example and whether you could copy something from it.

https://ui.shadcn.com/docs/components/skeleton

Image

4. **Style loading and no rows state**

Unfortunately, there is no indication in Figma on how these should be handled specifically for the tables. But maybe there are similar approaches documented that could be used here for consistency. Please let me know if you need direction on overriding these.

Image

Image

ObitoDevv:
## Description

- Component name: `NotificationGrid`
- Figma link or screenshot: [Figma Design](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=8446-188518&t=Z9yOThOzKCxx7IT2-4)

---

## General Checklist

---

## Related Issues

Fixes #134  (issue)

### DIFF
--- a/packages/ui-kit/lib/components/grids/notification-grid.tsx
+++ b/packages/ui-kit/lib/components/grids/notification-grid.tsx
@@ -2,7 +2,7 @@ import { notification } from '@maany_shr/e-class-models';
 import { AllCommunityModule, IRowNode, ModuleRegistry } from 'ag-grid-community';
 import { AgGridReact } from 'ag-grid-react';
 import { RefObject, useState, useCallback, useMemo, useEffect, useRef } from 'react';
-import { ChevronDown, ChevronUp, Search } from 'lucide-react';
+import { ChevronDown, ChevronUp } from 'lucide-react';
 import { BaseGrid } from './base-grid';
 import { formatDate } from '../../utils/format-utils';
 import { Button } from '../button';


packages/ui-kit/lib/components/grids/notification-grid.tsx
```diff
@@ -70,31 +123,157 @@ export const NotificationGrid = ({ notifications, onNotificationClick }: Notific
...
+            },
+            { all: 0, justdoad: 0, bewerbeag: 0, cms: 0 }
+        );
+    }, [modifiedNotifications, isLoading, notifications]);
+
+    // Client-side filtering logic for search
+    const doesExternalFilterPass = useCallback(
+        (node: any) => {
+            if (!node.data) return false;
+            const notification = node.data;
```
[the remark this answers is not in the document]
alebg: @ObitoDevv in general please don't use `any` when typing, nor the `as` whenever possible. Please either check how to type this properly, or give us a reason why this typing is not strict

packages/ui-kit/lib/components/grids/notification-grid.tsx
```diff
@@ -70,31 +123,157 @@ export const NotificationGrid = ({ notifications, onNotificationClick }: Notific
...
+
+
+    return (
+        <div className="flex flex-col h-full">
+            <Tabs.Root defaultTab="all" onValueChange={setSelectedTab}>
+                <TabList
+                    className="flex bg-base-neutral-800 rounded-medium gap-2 text-sm whitespace-nowrap min-w-max"
+                    variant="small"
+                >
+                    <TabTrigger value="all">
```
alebg, line 249: @ObitoDevv please don't hardcode anything in the ui kit, including platform names. If you need this information, pass it as props so the page can decide what to put as tab values

packages/ui-kit/lib/components/grids/notification-grid.tsx
```diff
@@ -70,31 +123,157 @@ export const NotificationGrid = ({ notifications, onNotificationClick }: Notific
...
+                        { 'All'} ({notificationCounts.all})
+                    </TabTrigger>
+                    <TabTrigger value="justdoad">
+                        { 'Just Do Ad'} ({notificationCounts.justdoad})
+                    </TabTrigger>
+                    <TabTrigger value="bewerbeag">
+                        { 'Bewerbeag'} ({notificationCounts.bewerbeag})
```
alebg, line 256: @ObitoDevv Please replace all "bewerbeag" for the correct name, which is "bewerbeagentur"

packages/ui-kit/lib/components/grids/notification-grid.tsx
alebg, line 249: @ObitoDevv this needs to variants as explained in the issue. in the CMS variant, we have the tabs with the names of the platforms. In the platform variant, we don't have the tabs so we just render all notifications passed to the grid.

packages/ui-kit/lib/components/grids/notification-grid.tsx
```diff
@@ -1,67 +1,120 @@
...
 import { BaseGrid } from './base-grid';
 import { formatDate } from '../../utils/format-utils';
 import { Button } from '../button';
+import { Tabs, TabList, TabTrigger, TabContent } from '../tabs/tab';
+import { getDictionary, isLocalAware } from '@maany_shr/e-class-translations';
 
-ModuleRegistry.registerModules([AllCommunityModule]);
+// Extend TNotification to include platform
+interface ExtendedNotification extends notification.TNotification {
+    platform?: string; // e.g., "justdoad", "bewerbeag", "cms"
```
alebg, line 14: @ObitoDevv @anilrise11 could you please avoid using so many optionals everywhere, in general? it gets hard to reason about (and hence review) if you do this. In this case I don't see why this would be optional, neither why would `notifications` below be optional.

Note that in the  platform (non-cms) variant of the grid, the ui kit doesn't need to know in which platform it is at all. It will just render all notifications passed to it

alebg (changes requested):
@ObitoDevv please address my inline comments

ObitoDevv:
Hi @alebg sir,

I’ve addressed all feedback. Here’s a brief summary of the changes:

1. Type Safety: Updated doesExternalFilterPass to use `RowNode<ExtendedNotification>` instead of any and typed `node.data` as `ExtendedNotification`.
2. Platform Names: Configured platform names to be passed via the platforms `prop`, removing hardcoded values.
3. Naming Correction: Replaced "`bewerbeag`" with "`bewerbeagentur`" across the component.
4. Variants: Implemented CMS variant with platform tabs and platform variant `(student/coach)` without tabs, rendering all notifications as passed.
5. Non-Optional Types: Ensured platform in `ExtendedNotification `and notifications in `NotificationGridProps `are non-optional for better type safety.
6. Platform Variant Logic: Ensured the platform variant renders all notifications without platform-specific filtering.

All changes are complete, and the component is ready for review. Please let me know if there’s anything else you’d like me to address!

alebg:
@ObitoDevv I also don't understand here if this PR is meant to be complete. Why do you have, for example, a `fetchNotificationsWithPlatform` mock function doing an important piece of logic in the official code? 

You don't need to filter anything with pure typescript before passing data to the grid. Instead, you can use the grid itself to filter by platform when you click on the tabs, on the cms variant.

ObitoDevv:
Hi @alebg sir, thank you for your feedback and for taking the time to review this PR!
I’d like to clarify a couple of points regarding the concerns you raised:

1. PR Completeness: This PR is not yet complete. My understanding was that the `usergrid PR` would be reviewed and merged first, and I planned to align this PR with the same structure and approach once that was finalized. Apologies for any confusion here—let me know if I should prioritize completing this PR independently or wait for the usergrid PR to be merged.

2. Filtering Logic in fetchNotificationsWithPlatform: Thanks for pointing out that the grid’s built-in filtering can handle platform-specific filtering via the CMS variant tabs. I have a few doubts about how the data is structured and how we should proceed:

- Data Source: Is the data coming from the backend already filtered by platform (e.g., separate API responses for each platform), or is it a mixed dataset containing a platform field? If it’s the latter, I assumed we’d need client-side filtering similar to what we did in the usergrid PR, where we filter users by role in the Tabs.Root component (see below for reference).
- If the backend sends pre-filtered data, I can remove the fetchNotificationsWithPlatform mock and related TypeScript filtering logic. Could you confirm the expected data structure?
- If the data is mixed, I was planning to leverage the grid’s tab-based filtering as shown in the usergrid code below, where we filter users `client-side` based on the selected tab (e.g., student, coach). Would this approach work for the `notifications grid`, or is there a different mechanism you recommend for the` CMS variant`?

```
<Tabs.Root defaultTab="all" onValueChange={handleTabChange}>
    <TabList className="flex bg-base-neutral-800 rounded-medium gap-2 text-sm whitespace-nowrap" variant='small'>
        <TabTrigger value="all" icon={<IconAll />} className="cursor-pointer">
            {dictionary.components.userGrid.all} ({userCounts.all})
        </TabTrigger>
        <TabTrigger value="student" icon={<RoleIcon role="student" />} className="cursor-pointer">
            {dictionary.components.userGrid.students} ({userCounts.student})
        </TabTrigger>
        ...
    </TabList>
    <div className="mt-4">
        <TabContent value="all" className="overflow-auto max-h-[70vh]">
            {renderGridWithActions(props.users || [])}
        </TabContent>
        <TabContent value="student" className="overflow-auto max-h-[70vh]">
            {renderGridWithActions(props.users?.filter(user => user.roles?.some(roleObj => roleObj.role === 'student')) || [])}
        </TabContent>
        ...
    </div>
</Tabs.Root>

```


I’ll work on removing the mock function and refactoring the logic once I have clarity on the data structure and your recommended approach. Could you also advise if there’s anything specific from the usergrid PR that I should replicate here to ensure consistency?

alebg (commented):
@MytsV I've done a cleanup of this PR, modified the following points:

- Show expand chevron only on truncated messages; works also when resizing the window
- Render filter buttons only on CMS variant
- Allow vertical auto-sizing, instead of setting an amount of rows
- Pass Mark All As Read callback from the outside
- Add 'New' column header for read/unread column, so users can sort on click
- Fix modal date filter
- Show notifications sorted by newer first on initial render
- Remove "label" default from TextInput
- Slightly increase width of filter modal to show all "Bewerbeagentur" string

It should be in line with the UserGrid that we merged already, but it will be good if you take a quick look.

MytsV:
@alebg all the little fixes and addressing the flickering is finished

PR #217, opened 2025-06-04
[TSK-1038] Assemble layout and home page with intermediary mocks

MytsV:
This PR contains tRPC mocks (without a required presenter or model structure) and assembly of home page and layout with all the needed fixes for ui-kit components. The placeholder e2e tests were deleted so that the actual application passes.


MytsV:
I will be fixing the CI issues

MytsV:
The build errors can be fixed by adding PLATFORM_ID environmental variable to the GitHub Actions setup, the other however are super weird and turned out not to be just local for me! Will try to figure them out

MytsV:
Fixed all the tests and linting, only adding platform id to environment is left for CI to pass

PR #229, opened 2025-06-25
[TSK-1070] Copy GitHub Actions setup for authentication to registries for private packages

MytsV:
This is required for the subsequent PRs that include CATS as a dependency.


alebg (approved):
Thanks a lot @MytsV !

PR #231, opened 2025-06-26
 Refactor File Upload Handling and Add File Metadata Model

jatinrise:
This PR introduces several improvements to the file upload functionality within the `drag-and-drop-uploader` component and related models:

1. i have used the new model(file-meta-data)  in the uploader section.
2. Changed the file buffer type in `FileUploadRequestSchema` from `Buffer` to `Uint8Array` to address compatibility issues, as `Buffer` is a Node-specific function not supported by browsers. This ensures the file data can be processed in a browser environment.

- __Note for Discussion__: I've used `Uint8Array` for the file buffer in the upload request. However, if the file isn't being manipulated on the client side beyond conversion to binary format, using the native `File` type might be a simpler and more performant option.


jatinrise:
Hello @alebg, I’ve incorporated all the requested changes—could you please review my PR when you have a moment?
I also had a quick question regarding the FileUploadRequestSchema implementation.

In this PR, the `FileUploadRequestSchema` uses `Uint8Array` for the file buffer, which involves converting the uploaded file to a binary format in the browser before upload. While this might be necessary for backend compatibility or client-side processing, I wanted to raise a point for discussion: if the file isn't being manipulated on the client side beyond this conversion, using the native `File` type could simplify the implementation.

- __Pros of switching to `File`__: Avoids unnecessary conversion to `Uint8Array` in the browser, potentially improving performance for large files, and keeps the code more aligned with the browser's native file handling.
- __Considerations__: We need to confirm if the backend API can handle `File` objects directly or if it requires binary data. If the latter, we might need to keep `Uint8Array` or handle the conversion closer to the upload point (e.g., in a service layer).

Could the team provide insight into the backend requirements for file uploads? Based on that, we can decide whether to update the schema to use `z.instanceof(File)` or retain the current approach with `Uint8Array`.

jatinrise:
Hello @maany, thank you for the feedback. I’ve incorporated all the suggested changes.
@alebg, could you please review?

alebg (approved):
Thanks a lot guys @jatinrise @anilrise11 !

I just made a small refactor, we won't use `lfn`s, but `ids` directly as needed. Also left some TODO notes in old components s.t we can remember to come back once we've cleaned up the way to perform uploads.

PR #239, opened 2025-07-02
[TSK-1069] [TSK-1095] Refactor pagination and error handling

MytsV:
This PR handles updates for pagination and errors:
- pagination now is an optional object and is included in each list request
- error components with the reusage of banners were implemented; the design of these should be adjusted
- not found component and loading component were implemented with approximate design, should be enhanced as well
- don't expect unauthenticated error from features that should be publically available; that's a kaboom
- don't show some components on errors instead of displaying a banner

Follow up:
- might logging of errors into browser console coming by default from presenter be required?


maany:
closing in favor of https://github.com/dream-aim-deliver/dad-e-class-monorepo/pull/249

PR #246, opened 2025-07-05
[TSK-1098] Supply contact email via an environmental variable

MytsV:
NEXT_PUBLIC_CONTACT_EMAIL now specifies a contact email, which will be displayed in default error messages.


maany (approved):
Contains duplicate commits/files from other PRs, may cause a conflict

PR #265, opened 2025-07-21
[TSK-1054] [TSK-1055] ui-kit: CoachNotes (create, edit & view)

### LINKED ISSUE #223: [TSK-1054] [TSK-1055] ui-kit: CoachNotes (create, edit & view)
### Screenshots:

- #### Coach Notes:

Image
[In figma.](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6913-268078&t=NwEOQISHDW7lZB5i-4)

---

- #### Coach notes empty state

Image
[In figma.](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6913-267111&t=NwEOQISHDW7lZB5i-4)

---
- #### Create Variant:

Image
[In figma.](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6913-292535&t=NwEOQISHDW7lZB5i-4)

---
- #### Edit Modal Variant:

Image
[In figma.](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6913-295907&t=NwEOQISHDW7lZB5i-4)

jatinrise:
This PR adds a CoachNotes component.

Screenshot 2025-07-21 at 15 52 07

fix #223


alebg (approved):
All good @jatinrise thanks a lot! Only thing is the placeholder in the input field which was optional. I've changed this and added a placeholder for the links' title and url in the dictionary

carlapbg (changes requested):
Hi, it looks good, thank you very much! The only thing missing is the edit modal variant. Please add it and let us know so we can review the complete component.

[CoachNotes (edit modal variant)](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6913-295907&t=TcpLyGJ9PmbC8BXf-4)

jatinrise:
@carlapbg @alebg, I've created a coachNotesEdit component. I have also created a Dialog component so that we can reuse it throughout the application.
Please review."

alebg:
Hi @jatinrise, thanks for the response.

So the props you set up are good, and they do allow for flexibility regardless if they are optional or not. It's just a matter of having an interface for your components that are easy to understand quickly. In general, setting a lot of optionals and defaults makes it less clear what should I pick in which situation. It's the same reasoning behind doing explicit variants with union types, we're forcing the developer using the component to make a deliberate choice, and if something doesn't work because of those choices we can know quicker where the problem comes from. Relying on defaults hides this complexity and then if something doesn't work properly, it's usually more time-consuming to trace back issues to the choice of defaults.

These things are important for us because these components will be used by multiple developers, possibly for years, and we want to minimize this kind of friction. In my opinion it is better to force deliberate choices, and then put examples (as you properly did in your stories). For example, if I (Luis) were to use your component right now for a page, and I find that your choice of defaults doesn't work for my usecase, I would need to dive into your component to find the cause. If you force me instead to pick explicit props, then it's on me to read your docstrings / check your stories, and figure out what I need.

Does this make sense to you?

PR #268, opened 2025-07-22
[TSK-1125] - ui-kit: CourseCardAddToPackage & List

carlapbg:
## Description

- Component name: `CourseCardAddToPackage`
- Figma link or screenshot: [Figma Design](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6385-175064&t=ZhQ9FnRDhLotQgwO-4)

---

## General Checklist

---

## Related Issues

Fixes # (issue)

### DIFF
--- a/packages/ui-kit/lib/index.ts
+++ b/packages/ui-kit/lib/index.ts
@@ -87,3 +87,5 @@ export { StudentCardList } from './components/student-card/student-card-list';
 export { YourStudentCard } from './components/student-card/your-student-card';
 export { YourStudentCardList } from './components/student-card/your-student-card-list';
 
+export { CourseCardAddToPackage } from './components/course-card/add-to-package/course-card-add-to-package';
+export { CourseCardAddToPackageList } from './components/course-card/add-to-package/course-card-add-to-package-list';


alebg (changes requested):
@carlapbg thanks a lot! So we're having a recurrent problem with the images some times. The error handling is ok, it's just that before that kicks in (e.g., if the network is slow), we see an ugly placeholder that makes the image container smaller than the later container with "Image not available". When there are many of these at the same time it looks a bit out of place.

immagine

In my opinion it's a good moment to do our own "Image" component that wraps around the `img` that you are using and it's being used all around the UI kit. The idea is to handle all of these things properly and in a single place, just like our Video player component does. Could you please create a task for yourself and do this? It doesn't have to be in this same PR, but I would also ask you later to refactor the ui-kit to change all `img` tags to the new `Image` component to standardize how we handle images. 


- Apart from this, for this specific PR, i didnt understand the stories, is the logic/naming flipped? for example, this one says "No courses included", but I understand that there are three courses included already?

immagine

(notice also here the placeholder before the images load)

PR #278, opened 2025-07-31
Fix: Use correct type TFileUploadRequest for file upload callbacks in LinkEdit component

### LINKED ISSUE #277: Fix: Use correct type TFileUploadRequest for file upload callbacks in LinkEdit component
## Problem
The `LinkEdit` component's `onImageChange` callback currently uses `TFileMetadata` type, which represents file metadata for files that already exist in the system. This is semantically incorrect for upload operations where we have:
- The actual `File` object from the input
- Basic metadata (name, size, etc.)

## Solution
Refactor the `onImageChange` callback to use `TFileUploadRequest` type instead, which is more appropriate for upload operations. The callback should return a `Promise<TFileMetadata>` to represent the uploaded file's final metadata.

jatinrise:
# Fix LinkEdit Component Type Issues and Improve UX

##  Issues Fixed

1. **Type Safety Issue**: `onImageChange` callback was incorrectly typed to accept `TFileMetadata` instead of `TFileUploadRequest`
2. **UX Bug**: "Add Link" button was showing when LinkEdit component was open, violating Figma design
3. **Logic Bug**: `!editingLinkIndex` condition incorrectly treated index `0` as falsy
4. **Upload Bug**: File uploads were automatically creating empty links in the array

## Changes Made

- Updated `onImageChange` interface to use `TFileUploadRequest` → `Promise<TFileMetadata>`
- Fixed semantic distinction between upload requests vs stored file metadata
- Updated all consuming components and stories to match new interface
- `Add Link Button`: Changed from `disabled` state to complete hiding during edit mode
 
fix #277

### DIFF
--- a/packages/ui-kit/lib/components/coach/coach-notes.tsx
+++ b/packages/ui-kit/lib/components/coach/coach-notes.tsx
@@ -280,7 +280,7 @@ function CoachNotesCreate({
                         )}
                     </div>
                 </div>
-                <div
+                {editingLinkIndex === null && <div
                     className="flex items-center mt-4 gap-2"
                     role="group"
                     aria-label="Add link divider"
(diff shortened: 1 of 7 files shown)


alebg (approved):
@jatinrise @anilrise11 thanks a lot for the proactiveness! looks good to me

PR #368, opened 2025-09-03
Edit course translationas de and en

bewerbeagentur:
Hi all,

I’ve prepared a short guideline document on terminology and wording in the frontend.
The main issue is that many labels are currently very course-focused, while in the frontend customers should see offers as the umbrella term (course, coaching, package, etc.).

To avoid confusion and keep the platform flexible, the file summarizes:

Why neutral wording matters in the frontend

Key principles (Offer vs. Course, etc.)

Terminology mapping (enroll/book, outline/offer details, etc.)

Language & style rules (Du-form in German, gender-neutral, consistency)

A request for you to check where labels are still too course-specific

📄 Please see the attached file in GitHub for details.

This is not only relevant for pages that are already visible, but also for the ones you are still developing or merging later on. Since many of these labels are likely set via defaults or pre-configured strings in the CMS/backend, it’s important that we align the terminology now. If this had been considered from the start, it would have been easier – but at this stage we can still make sure the platform feels consistent and clear for customers once everything goes live.

Thanks for reviewing this and making sure the wording is consistent across the frontend!
[Terminology_Wording_Frontend.pdf](https://github.com/user-attachments/files/22121014/Terminology_Wording_Frontend.pdf)


packages/ui-kit/stories/hero.stories.tsx
alebg, line 1: @bewerbeagentur just as a remark, this file doesn't affect the website. Anything ending in `.stories.tsx` is for the storybook preview, which is for development

maany:
Hi @bewerbeagentur, thanks for sharing the document and the guidelines regarding terminology mapping.

The labels (the keys that point to specific text) in the translation files are currently **course-focused**, because they reflect the course pages and components designed in Figma.  

Other types of offerings can have their own sections too. For example:  
- **Packages** already have a separate section.  
- **Masterclasses, workshops, or webinars** would also get their own sections when introduced.  

To keep the platform flexible, **translation files, UI components, pages, routes, and backend APIs are all designed to be extensible**. This means you can introduce new schemas for offerings such as workshops or masterclasses and then create custom cards or views that will show up in the dashboard—or any page you choose.  

For instance:  
- Keys like `beginCourse` or `courseTitle` are tied to the course experience (both the `/courses` page and the course backend).  
- If you want “Course Outline” to instead appear as **“Offer Details”**, you just update the translation value.  
- Similarly, you could rename “Package Detail” to “Offer Details” if that wording fits better.  

We treat each offering as its **own entity**. A course is different from a package, and you can already see this in the Figma designs:  
- A **course view** shows details such as duration and author.  
- A **package view** does not need those fields.  

Following the same principle, a **masterclass, webinar, hands-on-support, or even job-matching** (an idea from the design sprint) would all be modeled as independent entities.  

By treating each offering as a separate entity, the system remains **loosely coupled**. This means one offering can change without breaking the others, leaving room for future enhancements to be developed independently.  

---

### Why Loose Coupling Matters  

Loose coupling ensures the platform stays adaptable, cost-effective, and ready for growth.  

- **Enable unique features:** For example, a *masterclass* card might show an “Instructor Bio” and “Live Q&A link,” while a *course* card shows “Duration” and “Syllabus.” Each offering can have fields that make sense for it without forcing other offerings to change.  
- **Avoid ripple effects:** Changing how a package works won’t accidentally break the course pages.  
- **Prepare for future growth:** Adding a new type of offering—say a “Mentorship Program”—doesn’t require rebuilding the whole system.  
- **Keep costs down:** Because changes stay localized, you don’t need large-scale refactoring, which saves time and money.  
- **Stay flexible:** The platform can evolve with new business ideas or user needs, instead of being locked into one rigid model.  

By contrast, **tight coupling** would tie all offerings to a single structure. While that might seem simpler at first, it creates hidden dependencies. If you later add something like a masterclass, you’d have to rework core logic, translation files, and shared components—leading to higher development costs, and less flexibility to adapt.  

---

### A Visual Analogy: Lego vs. Superglue  

Think of each offering (course, masterclass, package, etc.) as a Lego block:  
- With **loose coupling**, every offering is its own Lego piece. You can snap on new pieces (new offerings) or swap them out without disturbing the rest of the structure. If you don’t like one piece, you simply replace it.  
- With **tight coupling**, all the Lego blocks are glued together. The structure looks stable, but if you want to change one piece, you have to break the glue, which often damages the rest of the structure. Making changes becomes slow, messy, and expensive.  

---

### Final Note

In the source code, we need to **optimize for Developer Experience and future costs of making changes in the codebase**. The translations file acts as a bridge between **User Experience (UX)** and **Developer Experience (DX)**.  

I want to reassure you that since your project design was frozen, keeping the project extensible has been our guiding principle throughout so far. Introducing tight coupling between offerings would be counterproductive to that goal.  

The similarities and distinctions between offerings (courses, packages, masterclasses, etc.) can and should be discussed during **design discussions**, but the **code should still follow the architectural principle of loose coupling**. That way, if the design evolves significantly in the future, we are not locked in or forced into costly refactoring.  

I hope this clarifies your concerns. We will anyway took a another look at the keys to see if they can be represented differently. However, that should not block you from assigning whatever values would make most sense to the users of the platform. :)

PR #371, opened 2025-09-03
Investigate why type-ignore comments were required in presenters

maany:


alebg:
Closing as this is not something causing any issues, hence not prioritized for launch.

Issue #376, opened 2025-09-04
[TSK-1312] Mock `Coaches tab (single course) - review`


alebg:
- Tab layout is already set up, you need to implement the specific "coaches" tab view
- This view is for coaches, course creators or admins, and comes in three variants (one per role)

## Features

- `ListCoaches`, [usecase models here](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/main/packages/models/src/usecase-models/list-coaches-usecase-models.ts), using the `courseSlug` request parameter

## Coach variant

[On figma](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=8453-252482&t=5eI1wuF6YJ2mqDXG-4). 

<img width="1631" height="1307" alt="Image" src="https://github.com/user-attachments/assets/60f7f31f-dcae-4384-a9eb-90209d70f0f0" />

- Note there's no filters, sorting, or manage buttons (remove/add coaches)

## Course creator variant

[On figma](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=6913-291104&t=5eI1wuF6YJ2mqDXG-4).

<img width="1452" height="1370" alt="Image" src="https://github.com/user-attachments/assets/0c231e1d-9e96-4f66-9767-38049e69b36d" />

- Has buttons to remove / add courses, and edit the course
- The add coach to course will call another backend endpoint (needs to be mocked): [usecase models here](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/main/packages/models/src/usecase-models/add-course-coach-usecase-models.ts)
- The remove coach from course will call another backend endpoint (also needs to be mocked): [usecase models here](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/main/packages/models/src/usecase-models/remove-course-coach-usecase-models.ts)
- No need to implement a filter modal here at all, just the fuzzy search and the sorting:

<img width="1176" height="74" alt="Image" src="https://github.com/user-attachments/assets/a232495c-d715-4999-bcf7-358e224c0d86" />

## Admin variant

- Same as above, but has an "Archive course" button. This will make a call to another backend endpoint (needs to be mocked): [usecase models here](https://github.com/dream-aim-deliver/dad-e-class-monorepo/blob/main/packages/models/src/usecase-models/archive-course-usecase-models.ts)

[...]

PR #397, opened 2025-09-10
[TSK-1342] [TSK-1343] feat(platform): add courses coach tab page

jatinrise:
This pr includes the following changes:

- Added `useCoachMutations` hook for managing coach additions and removals in courses.
- Created mock procedures for listing, adding, and removing course coaches.
- Developed presenters for handling responses from coach mutation operations.
- Introduced view models for add/remove coach operations to standardize response handling.
- Implemented a reusable `SearchInput` component for searching through coaches with fuzzy matching and debouncing.

Hello @alebg @MytsV,
I have created a page for the coach course tab pages. I have a small doubt regarding the presenter layer. For queries, we are creating a separate presenter layer, and for mutations like adding or removing a coach, I have created a new presenter layer. I would like to ask if this is a good approach, or should I use a single presenter layer for both. Please review my code and share your suggestions so that we can keep these points in mind going forward.

I have also found a bug in /courses/[slug] where I am getting all the roles, but it should only show admin and course creator roles. Could you please take a look at it?

feat #376

### DIFF
--- a/apps/platform/src/lib/infrastructure/client/pages/course/enrolled-course/enrolled-coaches.tsx
+++ b/apps/platform/src/lib/infrastructure/client/pages/course/enrolled-course/enrolled-coaches.tsx
@@ -22,7 +22,6 @@ import useClientSidePagination from '../../../utils/use-client-side-pagination';
 import MockTRPCClientProviders from '../../../trpc/mock-client-providers';
 import { useCoachMutations } from './hooks/use-coach-mutations';
 
-
 interface EnrolledCoachesProps {
     courseSlug: string;
     currentRole: string;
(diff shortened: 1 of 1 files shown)


alebg:
Hi @jatinrise, thanks a lot!

- Your two-presenter approach looks fine to me. It seems to be in line with what Viktoriia is doing (e.g., save course structure vs course structure presenters). @MytsV is this OK?

- If you mean that from the backend you're getting all roles, this is not a bug, this is the intended behaviour. Roles are *cumulative*, meaning: everyone gets `visitor`; `students` have also `visitor`; `coach` also has `student` *and* `visitor`, and so forth. This means that you always have to check for the precise role(s) you need by inclusion. The automated claude report above is saying precisely this ;) :

immagine

jatinrise:
Hello @alebg,
I think it’s not correct to filter out only admin and course_creator from roles because, if you see EnrolledCourse component inside `enrolled-course` file, it is showing the UI for the student role as well

`export default function EnrolledCourse(props: EnrolledCourseProps) {
    if (props.currentRole === 'student') {
        return <ProgressEnrolledCourse {...props} />;
    }
    return <EnrolledCourseContent {...props} />;
}`

Am i correct?

PR #420, opened 2025-09-16
[TSK-1327] Implement lesson progress connection

MytsV:


alebg (commented):
@MytsV thanks! Only remark is that your change in the domain service to use a full ...MappingContext for every partial function now forces every partial function to depend on all the fields of the context. This is not ideal because e.g., now lesson components that have nothing to do with video, need a video gateway. In particular, this now had me refactor a similar service I made for course material, and it forces me to give it the video gateway even if none of the material components include a video. No need to revert this back now, but please be mindful whenever you make this kind of changes to domain services

Issue #456, opened 2025-10-01
Implement `Course Completion (screen & review course)`


carlapbg:
[See this page in figma.](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=8454-304241&t=XehYNk8jrRIlP72F-4)
[And here.](https://www.figma.com/design/8KEwRuOoD5IgxTtFAtLlyS/Just_Do_Ad-1.2?node-id=8494-227061&t=XehYNk8jrRIlP72F-4)

Needs to be put in the course layout, student view. The modal is triggered when the student goes to the course layout and has the course status as completed (check with `getCourseStatus`).

## Course Completion Screen:

<img width="596" height="454" alt="Image" src="https://github.com/user-attachments/assets/2e07434d-47c2-455e-bd49-a089e43ff43a" />

## Review Course:
<img width="522" height="425" alt="Image" src="https://github.com/user-attachments/assets/dbaedb05-1146-4d36-8a7a-b6219d6a27fc" />

# Usecases:
All scaffolded.

### `getCourseStatus`

Scaffolded, but you can control a bit the response. If your course slug finishes with '-complete', then you will receive a course status completed. Else, you'll receive in progress with some progress number.

PR #481, opened 2025-10-10
Implement Course Completion (screen & review course)

jatinrise:
Hello,

I have raised the PR so that you can check the error I was encountering in the loop. I am currently unable to view my course as a student. I’ve already explained the issue in detail here:
🔗 [Issue #456 – Comment](https://github.com/dream-aim-deliver/dad-e-class-monorepo/issues/456#issuecomment-3387365720)

Currently, if the course is completed, a modal will pop up displaying the completion card and the feedback form.

I would like to request you to please test this functionality and share your feedback — whether it’s working fine or if you encounter any errors. Once the error is resolved, I will proceed to add the completion card to the layout.

Thank you!

### DIFF
--- a/packages/ui-kit/lib/utils/format-utils.ts
+++ b/packages/ui-kit/lib/utils/format-utils.ts
@@ -23,7 +23,7 @@ export function formatDate(date: Date = new Date()): string {
  * formatCertificateDate('2024-10-26T17:05:53.655296', 'de'); // "26. Oktober 2024"
  * formatCertificateDate(new Date(), 'en'); // "January 15, 2025"
  */
-export function formatCertificateDate(dateInput: string | Date, locale: string = 'en'): string {
+export function formatCertificateDate(dateInput: string | Date, locale = 'en'): string {
     try {
         let date: Date;
 


apps/platform/src/lib/infrastructure/client/pages/course-completion.tsx
```diff
@@ -97,6 +112,39 @@ export default function CourseCompletion({ slug }: CourseCompletionProps) {
         }
     }, [isLoggedIn, router]);
 
+    // Handle download certificate
+    const handleDownloadCertificate = () => {
+        if (certificateDataViewModel?.mode === 'default') {
```
alebg: @anilrise11 @jatinrise I don't understand the status of this PR. I thought I could check already the PDF certificate. 

You need to: (1) create a course; (2) log in with a different account with only 'student' (and 'visitor') role; (3) add in the database a course_student entry with the course ID and the student ID; (3) tweak a bit @jatinrise 's code to force the modal to show (since the backend is mocked). For me this works fine and I can see the modal:

immagine

I was trying to debug why the PDF doesn't generate, and it turns out is it because the `handleDownloadCertificate` function does nothing. Can we please know when this will be implemented? I don't see a reason to think this is blocked, we could already implement here the PDF generator.

alebg (changes requested):
@jatinrise @anilrise11 please check my inline comment

jatinrise:
Hello @alebg,

I have added the download certificate functionality, where I created a simple template. I’ve also incorporated some other changes. Please check it out so that we can make any necessary adjustments before merging this branch.
I have also added course-certificate storybook.

Thank you.

PR #494, opened 2025-10-15
Implement CMS Manage Homepage

jatinrise:
Hello,

I have created this PR as a draft so that you can check the issue. The `getHomePage` procedure is not working, which is why I have added mock data to implement the UI. It is currently showing the error in the screenshot below:

Screenshot 2025-10-15 at 22 31 21

As you know, this homepage has many image uploads. I have implemented the upload functionality, but there are some errors related to the procedure.

Additionally, please ignore build failures and UI-related issues for now. I will fix them after your approval of this PR.

### Some Recomdations

1. I think currently all the image types are just strings, which might not be correct. Please take a look at this.
2. If possible, could we integrate a tRPC playground? It would make it much easier for us to check and test every procedure before using it.

Thank you.

### DIFF
--- a/apps/cms/src/lib/infrastructure/client/pages/manage-home-page.tsx
+++ b/apps/cms/src/lib/infrastructure/client/pages/manage-home-page.tsx
@@ -55,6 +55,7 @@ export default function ManageHomepage() {
 
 	// Present data on mount and when response changes
 	useEffect(() => {
+		// @ts-ignore - Presenter doesn't handle progress states
 		homePagePresenter.present(getHomePageResponse, homePageViewModel);
 	}, [getHomePageResponse, homePagePresenter, homePageViewModel]);
 


alebg:
@jatinrise I've patched the backend, to give a clearer error whenever there is no HomePage object linked to the platform<>language, and also wired the get call properly here in the front. Now it should work. 

In your form, I saw it asks for a video ID as a plain number, but please do implement a video uploader like in the courses.

Also, please note my latest remarks (edits) in the issue for this task => on "Not Found" you should *not* fail but instantiate an empty form (like in the profile pages). In this way, when we start the app, we can still create a home page. This applies also to the offers and coaching page, all these 3 pages work very similarly.

With this code, I got your CMS form working and if you e.g., input some title and save, you will see it appear in the platform homepage.

@anilrise11 please prioritize this one together with #508 and #515 , if you get one working the other two will be almost the same.

PR #527, opened 2025-10-24
[TSK-1535] [TSK-1537] [TSK-1474] cms-ui: implement CMS Coaching Sessions page; Implement manage users page and send batch notification modal

jatinrise:
This PR adds the Coaching Session page to the CMS. I’ve added ` // @TODO` comments on the functions that handle clicks on the student, coach, and coaching session names. These need to be updated later with the actual domain URLs of the platform so that users can navigate to the correct pages.


alebg:
@anilrise11 @jatinrise just to understand, why did you change the ag grid tailwind styles?

alebg (approved):

PR #532, opened 2025-10-26
[TSK-1483] Implement package sale page

davidmarafuzgaspar:
Created the `package.tsx` page for the platform

Had to change the `BuyCompletePackageBanner` because it was not displaying the correct savings and price of package

### DIFF
--- a/apps/platform/src/lib/infrastructure/client/pages/package.tsx
+++ b/apps/platform/src/lib/infrastructure/client/pages/package.tsx
@@ -362,13 +362,13 @@ export default function Package({ locale, packageId }: PackageProps) {
                   imageUrl: course.imageUrl || '',
                   rating: course.averageRating,
                   duration: {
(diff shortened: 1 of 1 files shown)


alebg (changes requested):
Hey @davidmarafuzgaspar thanks a lot! So the course selector is not working as in the Figma:

immagine


After I exclude a course, it disappears so it can't be added again. Apart from that:

- Could you fix the TODOs please? we already have defined all routing in the app. If unsure of the route you need, please ask

- Could you please fix the duration of the course cards to render in format `HHhMM` instead of  the hours with decimals (which is not very natural)? I'm not sure where this is defined, but if you could please find where it's coming from and improve it that'd be ideal!

immagine


- I've added the course images to the response of the backend, you'll get it after (1) I make a backend release; (2) you rebase this branch to the updated code (later today)

PR #554, opened 2025-10-29
[TSK-1451] - Implement Your Reviews (coach workspace)

carlapbg:


apps/platform/src/lib/infrastructure/client/pages/workspace/your-reviews.tsx
```diff
@@ -0,0 +1,345 @@
...
+    DialogBody,
+    ReviewFilterModal,
+    ReviewFilterModel,
+    IconClose,
+} from '@maany_shr/e-class-ui-kit';
+import { useSession } from 'next-auth/react';
+import useClientSidePagination from '../../utils/use-client-side-pagination';
+
+interface YourReviewsProps {
+    roles: role.TRole[];
```
alebg: @carlapbg please use

`roles: TEClassRole[]`  and import `TEClassRole` from "@dream-aim-deliver/e-class-cms-rest" directly

alebg (changes requested):
@carlapbg very nice!! Please rebase this branch again to the latest main, and run `pnpm cache ...`. That sends a fix to the backend, and then if you populate some data, you can already see your page working end-to-end:

immagine

Just check the commits I made -- courses are optional for coaching sessions, so page and card (ui-kit) have to be able to handle that. I did a quick patch, but I leave it up to you to decide how it's handled best (visually).

Also if there's no review text the card looks a bit empty?

Apart from that, please address my inline-comments.

PR #571, opened 2025-11-03
fix(homepage-cms):fix the state on tab switching

jatinrise:
Hello @alebg,
Thank you for raising the issue. Yes, I also found it strange that switching tabs was causing a re-render. I discovered that React Query detects window focus and automatically refetches the data to ensure it's fresh — that’s why it was showing a stale state.

I've fixed the issue now.


jatinrise:
Hello @alebg, if you check the hook, it’s only handling the state and validating whether the state has changed or not. Yes, there is definitely room for improvement.

PR #578, opened 2025-11-04
[TSK-1532] cms-ui: implement @CMS Courses (Manage Courses; Platform specific)

jatinrise:


jatinrise:
Hello @alebg, I think we shouldn’t use this useEffect for the presenter to transform the response — it will always cause a re-render.

PR #693, opened 2026-06-01
Update navbar logo link to use dynamic URL based on locale

alebg:
This pull request updates the `Navbar` component to ensure that the logo link directs users to a locale-specific homepage URL instead of always linking to the root path. The main changes are:

Navigation link update:

* The `logoHref` variable is introduced to dynamically generate the homepage URL based on the current `locale`, ensuring users are directed to the correct language-specific site.
* The logo link's `href` attribute is updated to use the new `logoHref` variable, replacing the previous hardcoded root path (`"/"`).


maany:
Hi @alebg this is not the correct way to handle environment variables ( process.env). Please use env.ts and runtime config instead of `process.env.*` directly.

PR #706, opened 2026-08-03
refactor(platform/analytics): remove app-side Consent Mode bridge

### LINKED ISSUE #705: Remove app-side Consent Mode bridge — let the Usercentrics GTM template be the single owner
## Context

The consent work in #702 hardened the app's consent handling and added a live audit. While doing it, a structural question surfaced that is worth resolving separately: **most of the app's consent code exists to duplicate something the Usercentrics GTM template already does.**

A conventional CMP setup has *no* app-side consent code:

1. the CMP script loads first, blocking, in `<head>`
2. the CMP's GTM template sets Consent Mode `default = denied` and pushes `update` on the user's choice
3. all tags live in GTM and are consent-gated there

We deviate from that, and only part of the deviation is justified.

## What is actually necessary

**Gating first-party code on consent.** `OTelBrowserProvider` loads bundled JS — not a third-party `<script>` tag — so the Usercentrics autoblocker cannot reliably block it. Something in-app has to ask "did the user consent to this?" That is legitimate and normal.

## What is not

**The app-side Consent Mode bridge** (`consent-mode.ts` → `updateConsent()`), and the `analytics` / `marketing` / `preferences` category abstraction that exists to feed it.

Measured on production, both writers push to the same global Consent Mode state:

| Source | Payload |
| --- | --- |
| Usercentrics GTM template | 4-key: `ad_storage`, `ad_personalization`, `ad_user_data`, `analytics_storage` |
| our app | 5-key: the same four **plus** `personalization_storage` |

So the app duplicates all four core signals and adds exactly one the template does not send — `personalization_storage`, the signal Google gates the least on.

The cost of that duplication is concrete:

- it is what escalated the incident from `gcs=G101` to `gcs=G111` on eclass (the Wix site, with no app-side bridge, leaked only `G101`)
- two independent writers to one global state is a race/override class of bug
- the category flags are a lossy proxy that needed escape hatches: the Google-Analytics-under-`marketing` special case, and later per-service gating

## Proposal

**Remove the app-side Consent Mode bridge and the category abstraction. Keep per-service consent lookup.**

Remove:

- `consent-mode.ts` and `updateConsent()`
- `TConsentState.analytics` / `.marketing` / `.preferences`
- the `hasGoogleAnalyticsConsent` special case — it only exists because categories are a lossy proxy
- the category-mapping logic in `usercentrics-adapter.ts`

Keep:

- a small hook answering *"has the user consented to service X?"*, backed by `__ucCmp.getConsentDetails()` with the `EXPLICIT` check (`hasServiceConsent`)
- `OTelBrowserProvider` gating on its own CMP service
- the live audit suite (`apps/platform-e2e/src/consent-audit.spec.ts`)

Net effect is a deletion — likely more code removed than #702 adds.

## Sequencing

Do **not** do this at the same time as anything else. Suggested order:

1. **Merge #702** (hardening + audit). Already verified against production.
2. **Usercentrics category cleanup** — see the separate discussion: create a `statistics` category, move Google Analytics / Sentry / OpenTelemetry into it, move TikTok to `marketing`, decide on reCAPTCHA → `essential`. This independently removes the reason the GA special case exists.
3. **This issue** — delete the bridge and the category layer.

Doing 2 and 3 together would change consent behaviour and restructure the code simultaneously, with the live audit as the only safety net.

## Things to verify before removing

- **`personalization_storage` stops being sent entirely.** The UC template does not emit it. Confirm nothing depends on it — likely fine given how little Google gates on that signal, but it is a real behaviour change and should be a decision, not a side effect.
- **The denied baseline still holds.** No `consent default` command appears in the page dataLayer at all; the template applies it inside the GTM container on the Consent Initialization trigger. Removing the app bridge must not disturb that. The audit already asserts `gcs=G100` before interaction.
- **`track.*` events.** `track.ts` uses `sendGTMEvent` and does not gate on consent today; confirm the GTM-side consent checks cover it once the app stops pushing signals.
- **Any consumer of `useConsent()` reading category flags** must move to per-service lookups.

## Acceptance criteria

- `pnpm exec playwright test -c apps/platform-e2e/consent-audit.config.ts` passes on both production sites: nothing granted before interaction (`gcs=G100`), correct grants after Accept, and the granular case grants nothing beyond the accepted service.
- Exactly one writer to Consent Mode remains (the Usercentrics GTM template) — verifiable by there being a single `consent update` payload in the dataLayer rather than two.
- OpenTelemetry still starts only when its own CMP service is consented.

## Related

- #702 — consent-mode hardening + live audit (merge first)
- Usercentrics category cleanup (dashboard-side, no PR)

maany:
Closes #705.

The Usercentrics GTM template already owns Google Consent Mode. The app mirrored the CMP's category state into `gtag('consent','update',...)` as well, so one user decision produced two writes to one global state — and the app's was the less accurate of the two.

Measured on production, the two payloads were distinguishable by shape:

| Source | Payload |
| --- | --- |
| Usercentrics GTM template | 4 keys: `ad_storage`, `ad_user_data`, `ad_personalization`, `analytics_storage` |
| our app | the same 4 **plus** `personalization_storage` |

That duplication is what escalated a user who had accepted only Google Analytics from `gcs=G101` to `gcs=G111`. The template alone got the case right, which is how the app was identified as the source.

## Removed

- `consent-mode.ts` / `updateConsent()`, and the `useEffect` in `AnalyticsProvider` that called it
- `TConsentState.analytics` / `.marketing` / `.preferences`
- the category aggregation in `usercentrics-adapter.ts`, and with it the Google-Analytics-under-`marketing` special case — it only existed because categories were a lossy proxy

## Kept

- per-service consent: `hasServiceConsent()`, backed by `__ucCmp.getConsentDetails()` with the `EXPLICIT` check
- `OTelBrowserProvider` gating on its own CMP service

Gating bundled first-party JS is a real need the CMP autoblocker cannot serve. Mirroring Consent Mode was not.

`GtagBootstrapScript` keeps `window.dataLayer` / `window.gtag`. Nothing in the app calls `gtag` now, but it is the conventional global GTM expects, and dropping it can only be verified against a deployed environment — so it is deliberately left for a separate change rather than bundled here.

## Behaviour change

`personalization_storage` is **no longer sent at all**, since the Usercentrics template does not emit it. This was flagged in #705 as a decision rather than a side effect: Google gates the least on that signal, and having a second writer emit it was the anomaly.

## Size

| Area | Change |
| --- | --- |
| `apps/platform/src` | +100 / **−164** |
| `apps/platform/tests` | +316 / −345 |
| `apps/platform-e2e` | +90 (new single-writer audit test) |

## Verification

- `pnpm nx run platform:test` — 106 passing, 0 failing
- `eslint` across `apps/platform` — 0 errors

Two regressions are now pinned that were not before:

- **No Consent Mode command may reach the dataLayer on any path**, including the explicit-accept path — where writing one looks superficially correct. Both the unit repro and `AnalyticsProvider`'s own test assert this, each guarding against a vacuous pass by first confirming the app genuinely saw the grant.
- **Category assignment is fully ignored.** The adapter fixtures deliberately keep misleading categories (GA under `marketing`), so reintroducing any category read fails.

The live audit gains a single-writer check that fingerprints app-written updates by `personalization_storage` — the template never emits that key, so any update carrying it came from app code. It asserts on shape rather than a raw update count, since GTM may legitimately push more than one update.

## Sequencing note

#705 suggested doing the Usercentrics category cleanup (create `statistics`, move TikTok to `marketing`) before this. That ordering existed to give the GA special case a clean precondition — but this change deletes the category layer outright, so nothing here reads a category and the dependency does not apply. The dashboard cleanup remains worth doing on its own merits (user transparency: today a visitor must accept **Marketing** to allow analytics) and is unaffected by this PR.

## Before merging

The single-writer audit test necessarily runs against **deployed** code — it cannot pass before release. Run it right after deploy:

```
pnpm exec playwright test -c apps/platform-e2e/consent-audit.config.ts
```


maany:
Thanks — the finding was real and is fixed in `ebe90384`. One correction on the diagnosis, and a note on something the review missed.

### Valid: stale properties in `otel-browser-provider.test.tsx`

Confirmed and fixed. I updated the `state()` helper but not its call sites. Removed from all three (the review spotted two; `:139` had a stale `analytics: false` as well).

One of them was worth more than a mechanical deletion. `:160` read:

```ts
const { adapter } = controllableAdapter(state({ analytics: true }));
```

under the title *"treats an absent consent record as no consent"*. Since `state()` returns `{ services: {} }`, that was modelling an **empty** record — "the CMP was asked and granted nothing" — not an absent one. Those are different states, and only the absent one is what the test name claims. It now passes `{}` (no `services` key at all), which is the case `hasServiceConsent()` actually guards.

### Correction: the typecheck failure was real, but not this

The review's reasoning — Vitest doesn't typecheck, so tests can pass while `tsc` fails — was right, and `main` did fail. But not on excess-property checks: `Partial<TConsentState>` accepts those literals without complaint, because excess-property checking is suppressed for a type with an optional index-free shape when the value flows through a variable... more simply, these object literals **do** get flagged in principle, but the build never reached them. `apps/platform/tests` is outside the `next build` type-check scope.

The actual failure was in `gtag-bootstrap-script.tsx`, in code the review passed as "reasonable scope boundary":

```
Type error: Subsequent property declarations must have the same type.
  Property 'dataLayer' must be of type 'Object[] | undefined',
  but here has type 'unknown[] | undefined'.
```

When I moved the `declare global` block out of the deleted `consent-mode.ts`, I carried `dataLayer?: unknown[]` with it. `@next/third-parties` already declares `Window.dataLayer` globally as `Object[]`, and two global declarations of one property must agree. Only `window.gtag` needed declaring here; `dataLayer` is now left to the library, with a comment saying why so it doesn't get re-added.

### Process note

This is on me: I reported "106 passing, 0 lint errors" without a green build. My local `tsc -p apps/platform` had failed on unbuilt project references (TS6305) and I moved past it instead of getting a working typecheck — so I shipped a claim I hadn't actually verified. Now confirmed properly:

- `pnpm nx run platform:build` — **succeeds** (this is the check that catches the above)
- `pnpm nx run platform:test` — succeeds
- `eslint apps/platform` — 0 errors

### On the nit

Agreed, and the duplication is deliberate. `analytics-provider.tsx`, `gtag-bootstrap-script.tsx` and `usercentrics-adapter.ts` are the three places someone would plausibly reintroduce a consent write, and the invariant is only useful where it would be violated. Leaving as-is.

maany:
**Post-deploy verification: all 7 live audit checks pass on both production sites.**

The two that verify this change specifically:

**Single writer** — Accept All on eclass produced exactly **one** consent update:

```json
{"ad_storage":"granted","ad_personalization":"granted","ad_user_data":"granted","analytics_storage":"granted"}
```

Four keys, from the Usercentrics GTM template. No `personalization_storage` — the fingerprint of an app-side write. There were two writers before.

**Granular save** — accepting only Google Analytics now yields one update with `analytics_storage: granted` and **every advertising signal denied**, plus zero OpenTelemetry collector posts. That is the exact case that reported `gcs=G111` and sent telemetry for an explicitly refused service.

Both sites remain `ALL_DENIED` / `gcs=G100` before the banner is touched.

Artifacts are in #707 (evidence only, no source changes) — this PR was merged before the audit could run, since it needs deployed code.

